PR #514, opened 2026-04-12
Fix transform-origin presentation attribute and single-keyword parsing

jwmcglynn:
## Summary

Fixes `transform-origin` as an SVG presentation attribute (issue S2 from proposed_issues_2026q2.md). Two bugs:

- **Presentation attribute not registered**: `transform-origin` was missing from `kValidPresentationAttributeEntries`, so `<rect transform-origin="50% 50%">` was silently ignored (only `style="transform-origin: ..."` worked).
- **Single-keyword parsing broken**: `ParseTransformOrigin` required whitespace + a second coordinate after the first, so `transform-origin: center` errored out. Now handles single-keyword syntax per CSS spec (`center` → 50% 50%, `top` → 50% 0%, etc.).

Enables **20 resvg tests** in `structure/transform-origin/*`.

## Test plan

- [ ] CI passes (linux, macOS)
- [ ] 20 transform-origin resvg tests pass with default threshold
- [ ] Existing transform-origin style="" usage still works (no regression)

### DIFF
--- a/donner/svg/properties/PropertyRegistry.cc
+++ b/donner/svg/properties/PropertyRegistry.cc
@@ -648,31 +648,57 @@ ParseResult<TransformOrigin> ParseTransformOrigin(std::span<const css::Component
     return parser::ParseLengthPercentage(component, true);
   };
 
+  // Helper to detect if the first token is a Y-axis keyword (top/bottom).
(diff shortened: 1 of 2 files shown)


jwmcglynn (commented):
Approved

PR #520, opened 2026-04-12
Build: switch from monolithic entt.hpp to modular EnTT headers

jwmcglynn:
## Summary

Replace the 95,903-line monolithic `<entt/entt.hpp>` single-include header with targeted modular includes (~1,960 lines total). Donner only uses `entt::entity`, `entt::basic_registry`, and `entt::basic_handle`.

**Changes:**
- `EcsRegistry.h`: 4 modular headers instead of `entt.hpp`
- New `EcsRegistry_fwd.h`: lightweight forward-declaration header for 15 headers that only need type names
- Fix missing `<map>` includes in 8 files (was transitively included via `entt.hpp`)
- Fix missing `entt/entity/helper.hpp` in 3 files using `entt::to_entity()`
- 40 files changed total

**Expected impact:** 30-50% faster compilation for the 48 translation units that transitively include ECS headers.

## Test plan

- [ ] `bazel build //donner/...` passes (verified locally, 324 non-renderer targets clean)
- [ ] CI passes on all jobs
- [ ] No functional changes — purely a header reorganization

### DIFF
--- a/donner/base/BUILD.bazel
+++ b/donner/base/BUILD.bazel
@@ -16,6 +16,7 @@ donner_cc_library(
         "ChunkedString.h",
         "CompileTimeMap.h",
         "EcsRegistry.h",
+        "EcsRegistry_fwd.h",
         "FileOffset.h",
         "Length.h",
         "MathUtils.h",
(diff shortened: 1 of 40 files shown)


jwmcglynn (commented):
Approved

PR #545, opened 2026-04-19
Bazel: hermetic Linux sysroot for remote execution

jwmcglynn:
## Summary

Unblock Bazel remote execution (`--config=re`) against bazel-re1 end-to-end. Four fixes land together because neither the sysroot nor the follow-ups are individually sufficient — RE was hitting a different wall at each layer (compile → link → py_binary → emcc).

1. **Hermetic Linux sysroot.** Wire Chromium's prebuilt Debian Bullseye x86_64/aarch64 sysroots into `llvm_toolchain` via `llvm.sysroot(...)` so libc++'s `<wchar.h>` can find glibc's `mbstate_t`. Extension keys the tarballs under `non_bcr_deps` (dev-only, BCR consumers never see them). `http_archive` uses `type = "tar.xz"` because the GCS URLs are bare-sha256 paths with no extension. Filegroup excludes `etc/`, `lib/systemd/`, `var/`, and a few `usr/share/*` subtrees — systemd units use escape-sequence filenames that Bazel can't represent as target names.
2. **libc++ `copy_file_range` shim.** Chromium's bullseye_arm64 sysroot exports `copy_file_range@GLIBC_2.27` as a *non*-default-versioned symbol (single `@`, not `@@`), so unversioned refs from `toolchains_llvm` 21.1.6's prebuilt libc++.a (in `std::__fs::filesystem::detail::copy_file_impl`) don't auto-resolve. Scanned libc++.a's 457 undefined refs against the sysroot's 70 non-default glibc symbols — `copy_file_range` is the only intersection, so a single-symbol shim suffices. `//third_party/libc_compat` provides an unversioned `copy_file_range` that forwards to the versioned glibc via `.symver`, injected as an implicit dep by `donner_cc_binary/test/fuzzer` via `libc_compat_deps()` selected on `//build_defs:llvm_latest_linux`. Keeps LLVM at 21.1.6.
3. **Hermetic Python toolchain.** Register a `rules_python` Python 3.12 toolchain in `MODULE.bazel` and set `bootstrap_impl=script` under `--config=latest_llvm` so py_binary's stage-1 launcher invokes the hermetic interpreter through the runfiles tree instead of falling through the `#!/usr/bin/env python3` shebang. Without this, `//tools:generate_embedded_test_resources` fails on the RE sandbox with `env: 'python3': Permission denied`.
4. **Emscripten/V8 jitless.** NativeLink's RE sandbox forbids W+X memory pages, so `emcc`'s bundled Node.js V8 crashes with SIGTRAP in `OS::SetPermissions` the moment it tries to baseline-compile `compiler.mjs`. `--action_env=NODE_OPTIONS=--jitless` keeps Node in interpreter-only mode; emscripten is happy with it.

## Sysroot provenance

| arch | sha256 | size |
| --- | --- | --- |
| `debian_bullseye_amd64_sysroot.tar.xz` | `52d61d4446ffebfaa3dda2cd02da4ab4876ff237853f46d273e7f9b666652e1d` | ~19.7 MB |
| `debian_bullseye_arm64_sysroot.tar.xz` | `c7176a4c7aacbf46bda58a029f39f79a68008d3dee6518f154dcf5161a5486d8` | ~18.4 MB |

URL = `https://commondatastorage.googleapis.com/chrome-linux-sysroot/<sha256>`, per Chromium's [`build/linux/sysroot_scripts/install-sysroot.py`](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/build/linux/sysroot_scripts/install-sysroot.py).

## Test plan

- [x] `bazel build --config=latest_llvm //donner/base:base` — local hermetic LLVM build works with the new sysroot (~243 s cold).
- [x] `bazel test --config=latest_llvm //donner/base:base_tests` — local test binary (uses `std::filesystem`) links *and* passes at runtime, proving the libc++ shim forwards correctly.
- [x] `bazel build --config=re //donner/base:base` — RE leaf build, warm 14 s.
- [x] `bazel build --config=re //...` — **full validation, green**, against `grpcs://192.168.1.83:50051`:
      ```
      INFO: Elapsed time: 1923.516s, Critical Path: 327.92s
      INFO: 4798 processes: 6602 action cache hit, 2 disk cache hit,
        1 remote cache hit, 826 internal, 3969 remote.
      INFO: Build completed successfully, 4798 total actions
      ```
      ~32 min wall, 3,969 remote executions on bazel-re1.

🤖 Authored with Claude Code.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -54,6 +54,17 @@ common:latest_llvm --extra_toolchains=@llvm_toolchain//:all
 common:latest_llvm --//build_defs:llvm_latest=1
 # Disable warning about not eliding copy on return (NRVO), which started triggering in skia on LLVM 21
 build:latest_llvm --copt=-Wno-nrvo
+# Use rules_python's hermetic Python launcher rather than the default
+# `#!/usr/bin/env python3` shebang. Without this, py_binary targets still rely
+# on a host python3 in PATH, which the Bazel remote-execution sandbox
+# (bazel-re1, NixOS) does not provide.
+common:latest_llvm --@rules_python//python/config_settings:bootstrap_impl=script
+# Disable V8's JIT in the emscripten Node runtime. bazel-re1's NativeLink
+# sandbox forbids creating executable memory pages (W+X), so V8 aborts with
+# SIGTRAP in `OS::SetPermissions` the moment it tries to baseline-compile
+# anything. `--jitless` keeps Node in the interpreter-only mode, which
+# emscripten's `compiler.mjs` is happy with.
+common:latest_llvm --action_env=NODE_OPTIONS=--jitless
 
 # Renderer backend selection
 common:skia --//donner/svg/renderer:renderer_backend=skia
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -50,6 +50,8 @@ use_repo(
     "harfbuzz",
     "resvg-test-suite",
     "skia",
+    "sysroot_linux_aarch64",
+    "sysroot_linux_x86_64",
     "tracy",
     "wgpu_native_linux_aarch64",
     "wgpu_native_linux_x86_64",
@@ -157,6 +159,20 @@ llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", 
 llvm.toolchain(
     name = "llvm_toolchain",
     llvm_version = "21.1.6",
+)
+
+# Hermetic Linux sysroots: keep the toolchain self-contained on Bazel
+# remote-execution workers (bazel-re1, NixOS) that lack system glibc
+# headers. See third_party/bazel/non_bcr_deps.bzl for the tarballs.
+llvm.sysroot(
+    name = "llvm_toolchain",
+    targets = ["linux-aarch64"],
+    label = "@sysroot_linux_aarch64//:sysroot",
+)
+llvm.sysroot(
+    name = "llvm_toolchain",
+    targets = ["linux-x86_64"],
+    label = "@sysroot_linux_x86_64//:sysroot",
 )
 use_repo(llvm, "llvm_toolchain", "llvm_toolchain_llvm")
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -142,6 +142,21 @@ new_local_repository(
 
 bazel_dep(name = "rules_python", version = "1.9.0", dev_dependency = True)
 
+# Hermetic Python toolchain. Without this, rules_python's py_binary targets
+# fall back to `/usr/bin/env python3` on the host — which is fine for local
+# dev on dev1, but on the Bazel remote-execution sandbox (NixOS) no such
+# host interpreter is available. Registering `python.toolchain()` makes
+# rules_python ship a prebuilt CPython through the action cache so RE
+# workers execute the same interpreter as local builds.
+python = use_extension(
+    "@rules_python//python/extensions:python.bzl",
+    "python",
+    dev_dependency = True,
+)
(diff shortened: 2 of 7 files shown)


jwmcglynn:
🤖 Update from RE validation after the `libxml2` worker patch landed.

## What now works

- `bazel build --config=re //donner/base:base` — remote-executes end-to-end (4 remote compiles, 1 remote link) against `grpcs://192.168.1.83:50051` in ~14 s warm.
- Partial `//donner/base/... //donner/css/... //donner/svg/core/...` — 107 remote actions, all compiles succeed; only the std::filesystem-using test targets fail at link (see below).
- Partial `//...` — 7,120 actions complete, **1,167 remote executions succeed** (compiles + genrules + links). Wall time to first non-sysroot failure: **9 m 12 s**.

## New worker-side gap: Python3 on bazel-re1

`/tools:generate_embedded_test_resources` is a `py_binary` genrule; it fails on the RE sandbox with:

```
env: 'python3': Permission denied
```

The sysroot fix doesn't reach into the Python toolchain — Bazel's autodetect strategy is falling back to a host `python3` the RE sandbox can't exec. Two options for the operator:

1. Make `python3` available to the bazel-re1 sandbox user (same shape as the libxml2 fix), or
2. Register a hermetic Python toolchain in donner (`python.toolchain(python_version="3.12")` via `rules_python`), which would ship an interpreter through the action cache. This is more work but removes the last PATH dependency.

## New toolchain/sysroot mismatch: libc++.a copy_file_range (aarch64)

When linking targets that pull in `std::filesystem`, ld.lld fails:

```
ld.lld: error: undefined symbol: copy_file_range
>>> referenced by operations.cpp
>>>               operations.cpp.o:(... copy_file_impl ...)
>>>               in archive .../lib/aarch64-unknown-linux-gnu/libc++.a
>>> did you mean: copy_file_range@GLIBC_2.27
>>> defined in: sysroot/lib/aarch64-linux-gnu/libc.so.6
```

Diagnosis: the prebuilt libc++.a in `toolchains_llvm` 21.1.6 (aarch64) was built against a libc where `copy_file_range` was unversioned, but Chromium's Debian Bullseye arm64 sysroot exports it as `copy_file_range@GLIBC_2.27` (single-@, not default). The two don't link.

This is an orthogonal issue to the sysroot PR — same failure happens regardless of whether the sysroot lives in `non_bcr_deps` or anywhere else — so merging this PR is still the right move; the `std::filesystem` mismatch is a follow-up. Likely fixes, from easiest to hardest:

1. Swap the prebuilt LLVM toolchain for one built against a matching glibc sysroot (ask upstream `toolchains_llvm` whether the `-linux-gnu` aarch64 triple is available with a Bullseye-sysrooted libc++).
2. Bump to an LLVM version whose libc++ uses `copy_file_range@GLIBC_2.27` directly (`.symver` hint in the headers it was compiled with).
3. Build libc++ from source against the sysroot inside our build (big change).

For now, RE works for everything that isn't `std::filesystem` and isn't a `py_binary`. That's enough to cut the time on compile-bound top-of-tree builds dramatically.

jwmcglynn:
🤖 PR scope expanded and validated end-to-end.

All four RE blockers are now addressed in this branch; full `bazel build --config=re //...` against `grpcs://192.168.1.83:50051` completes successfully (4798 actions, 3969 remote executions, 32 m 4 s wall). LLVM stays at 21.1.6.

See the updated PR description for the full breakdown; commit summary:

1. `edeb4f0` — hermetic sysroot (original commit).
2. `15fddd2` — libc++ `copy_file_range` shim + hermetic `rules_python` toolchain + `NODE_OPTIONS=--jitless` for emscripten.

Ready for review.

PR #553, opened 2026-04-20
Bazel: macOS remote execution toolchain wiring

jwmcglynn:
## Summary

Revives the macOS remote-execution branch on top of current `main` and keeps the diff focused on the Bazel wiring needed for `klein` -> `bazel-re2`.

### Changes

- Adds named Linux/macOS target platforms used by local and remote Bazel configs.
- Splits the LLVM toolchain into Linux exec-platform variants so a macOS client does not upload a Mach-O clang to Linux workers.
- Leaves macOS exec on `apple_support` / Xcode clang so `objc_library` targets such as GLFW Cocoa continue to build.
- Propagates `DEVELOPER_DIR` and `SDKROOT` under `--config=re` only. Local macOS sandboxed builds already inject these, and setting them globally trips Bazel's Darwin env rewriting.
- Patches emsdk `env.sh` so the hermetic Python runtime can find `libpython` under macOS remote execution.
- Cleans up the old wordy comments and drops the stale generated-CMake conflict churn during rebase.

### Verified

- `git diff --check`
- `./tools/llm-bazel-wrap.sh build --config=latest_llvm //donner/editor --jobs=8 --verbose_failures`
- `./tools/llm-bazel-wrap.sh build --config=re //donner/editor --jobs=8 --verbose_failures`
- `./tools/llm-bazel-wrap.sh build --config=re //... --jobs=1 --verbose_failures`
  - 1154 targets
  - 6371 total actions
  - 2301 remote actions

### Operational note

The current single `bazel-re2` worker is not stable at high client concurrency for the full graph. `//donner/editor` passed at `--jobs=8`, but broader `//...` runs saw remote clang exit 9 at `--jobs=8` and `--jobs=4`; `--jobs=1` completed cleanly. We should tune the macOS worker/client concurrency before making a high `--jobs` default for full-graph RE builds.


build_defs/rules.bzl
```diff
@@ -54,20 +54,12 @@ def _banned_patterns_lint_test(name, srcs, hdrs, tags = [], **_kwargs):
 
 def llvm21_macos_workaround_linkopts():
     """
-    Returns linkopts needed for LLVM 21 __hash_memory symbol workaround on macOS.
+    No-op retained for call-site compatibility.
```
jwmcglynn: Do fuzzers still work?  If so - remove this entirely instead of maintaining a vestigial stub

build_defs/rules.bzl
```diff
@@ -112,13 +104,12 @@ def libc_compat_deps():
 def fuzzer_compatible_with():
     """
     Returns a list of labels that the fuzzer rules are compatible with.
-    """
 
-    # Only run on Linux, or if --config=toolchains_llvm is used.
-    # Since the macOS clang is missing libclang_rt.fuzzer_osx.a, we cannot use the built-in macOS toolchain
+    Fuzzers are Linux-only because macOS exec uses Xcode clang, which does not
```
jwmcglynn: This change was *specifically* to enable fuzzers on macOS.  Do not remove this if it breaks fuzzers

PR #554, opened 2026-04-20
Geode perf: M2 path-encode cache + M4.2 transient-texture pool

jwmcglynn:
## Summary

Implements two cross-cutting Geode performance milestones from design doc
0030 so idle re-renders of the same document allocate zero CPU-side
encodes *and* zero transient textures:

- **M2: `GeodePathCacheComponent`** — caches `GeodePathEncoder::encode`
  (cubic→quadratic, monotonic split, band decomposition) and
  `Path::strokeToFill` on a per-entity ECS component, keyed by entt
  `on_update<ComputedPathComponent>` invalidation.
- **M4.2: transient-texture pool** — exact-size
  `(width, height, format, sampleCount, usage)` free list covering
  isolated-layer / filter-layer / mask / clip-mask scratch textures
  plus the mix-blend-mode `dstSnapshot`. Releases are deferred to
  after `queue.submit` so GPU work finishes before the same texture
  is handed back out.

## Headline counter deltas (frame 2, same document)

| Fixture               | pathEncodes | textureCreates |
|-----------------------|-------------|----------------|
| SimpleShapes          | 3 → 0       | 2 → 0          |
| Moderate (w/ layer)   | 2 → 0       | 8 → 0          |
| lion.svg              | 132 → 0     | 2 → 0          |
| Ghostscript_Tiger.svg | 305 → 0     | 2 → 0          |

Both invariants are asserted by
`{SimpleShapes,Moderate,Lion,GhostscriptTiger}_NoDirtyPath_{ZeroEncodes,ZeroTextures}`
in `GeodePerf_tests.cc`.

## What's in the stack

Commits (in order):

1. `design: retain M1.f.2 investigation notes from abandoned attempt`
2. `design + tests: M2 GeodePathCacheComponent — red repro tests first`
3. `ShapeSystem: gate ComputedPathComponent writes on content change`
4. `RendererInterface: add sourceEntity field to PathShape`
5. `Geode M2: GeodePathCacheComponent — zero-encode steady state`
6. `Geode M2: don't disconnect entt listener at renderer teardown`
7. `M4.2: red repro tests — textureCreates == 0 on repeat render`
8. `M4.2 step 1: share dummy textures across GeoEncoders via GeodeDevice`
9. `M4.2 step 2: TexturePool for layer/filter/mask/clip-mask scratch`
10. `M4.2 step 3: tighten baseline ceilings + mark design doc landed`
11. `CMake mirror: skip geode_path_cache_component in non-geode targets`

## Design notes worth flagging at review

- **Entity plumbing.** `PathShape` gained a `sourceEntity` field instead of
  widening `RendererInterface::drawPath`. Non-driver callers (overlay,
  tests) leave it null; Geode keys its cache on it.
- **ShapeSystem content-equality gate.** New `Path::operator==` +
  `emplaceComputedPathIfChanged` helper. This is the invalidation
  edge M2 relies on: `on_update<ComputedPathComponent>` only fires
  when the spline actually changed. Without this, the cache would
  be wiped on every frame where any entity was dirty.
- **Listener teardown is intentionally not disconnecting.** Fixed a
  UB crash (`RendererPublicApiTest.StrokeDasharrayDiffersFromSolid`)
  where `SVGDocument` destructs before `Renderer` in the same scope
  and the Impl's `~` called `disconnect<>()` on a freed registry.
  Listeners are free functions with no `this`-capture, so leaving
  them attached is safe — they die with the registry.
- **M4.2 uses exact-size keying, not power-of-two bucketing.** Simpler
  and sufficient for the repeat-same-size target; viewport-resize
  scenarios will miss all buckets on the new size. Bucketing is
  listed as a follow-up in design doc 0030.

## Deferred to follow-up PRs

- GPU-resident vertex/band/curve arena handle retention on
  `GeodePathCacheComponent` (cache the upload, not just the CPU
  bytes).
- `<use>` instancing via the cache (M6).
- Power-of-two size bucketing for viewport-resize scenarios (M4.2
  extension).
- `GeodeFilterEngine`-internal texture pool + command encoder
  sharing (M5).

## Test plan

- [x] `//donner/svg/renderer/geode:geode_perf_tests` — all 13 perf
      assertions green (4 M2 zero-encode + 4 M4.2 zero-textures + 4
      tightened baselines + counters-reset).
- [x] `//donner/svg/renderer/tests:renderer_geode_golden_tests`.
- [x] `//donner/svg/renderer/tests:renderer_public_api_tests` (catches
      the pre-fix StrokeDasharray crash).
- [x] `//donner/svg/renderer/tests:renderer_tests` (tiny-skia default backend).
- [x] `//donner/svg/renderer/tests:resvg_test_suite_geode` — 0 regressions.
- [x] `//donner/svg/renderer/...` — 69 pass / 3 skipped.
- [x] `//donner/base:base_tests`, `//donner/svg/components/shape:shape_system_tests`.
- [x] `tools/presubmit.sh --fast` (CMake mirror validation + clang-format).
- [ ] Let CI run the full `bazel test //...` + fuzzers + Linux build.

### DIFF
--- a/docs/design_docs/0030-geode_performance.md
+++ b/docs/design_docs/0030-geode_performance.md
@@ -170,8 +170,13 @@ algorithm.
   - [x] Tightened Milestone 0 counter ceilings in `GeodePerf_tests.cc`:
     Lion `bufferCreates` ceiling now 10 (down from 800); Lion
     `bindgroupCreates` ceiling stays 200 pending M1.f.2.
-- [ ] Milestone 2: `GeodePathCacheComponent` — cache the CPU encode
-  (Tier 3 findings, maps to 0017 Phase 5 bullet 1).
+- [x] Milestone 2: `GeodePathCacheComponent` — cache the CPU encode
+  (Tier 3 findings, maps to 0017 Phase 5 bullet 1). _Landed
+  2026-04-20._ Observed frame-2 `pathEncodes` deltas:
+  SimpleShapes 3→0, Moderate 2→0, Lion 132→0,
+  Ghostscript_Tiger 305→0. GPU-arena-handle retention (original
+  bullet 3) deferred to a follow-up PR; `<use>` instancing
+  stays on Milestone 6.
   - Cache only the final `EncodedPath`, not the intermediate
     `quadPath` / `monotonicPath`. The intermediate stages are fused
     inside `GeodePathEncoder::encode` and retaining them buys
--- a/docs/design_docs/0030-geode_performance.md
+++ b/docs/design_docs/0030-geode_performance.md
@@ -305,20 +305,34 @@ algorithm.
   - [x] Counter delta: Moderate (1 `<g opacity>` layer) drops from
     `submits=4` → `submits=2` (frame + readback). Ceiling tightened
     in `GeodePerf_tests.cc`.
-- [ ] Milestone 4: Transient render-target pool (Tier 2 findings).
+- [x] Milestone 4: Transient render-target pool (Tier 2 findings).
+  _Landed 2026-04-20 (M4.2)._
   - [x] M4.1: Reuse `impl_->target` and `impl_->msaaTarget` across frames
     when `pixelWidth`/`pixelHeight` are unchanged. _Landed 2026-04-19._
     Tracks `impl_->targetWidth`/`impl_->targetHeight` in Impl; beginFrame
     only recreates when the new viewport size differs. The
     `CountersResetBetweenFrames` test gates this with
     `EXPECT_LT(secondCounters.textureCreates, firstCounters.textureCreates)`.
-  - [ ] Size-bucketed free list for layer/filter/mask scratch textures:
-    round (w,h) up to the next power of two per axis, key on
-    `(bucket_w, bucket_h, format, sampleCount)`, recycle on pop/endFrame.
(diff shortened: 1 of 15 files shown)


jwmcglynn:
🤖 Addressed both Codex suggestions in e1c7c5e4:

**P1 — registry-pointer-reuse**: replaced the `&registry == cacheInvalidationRegistry` pointer guard with a `ListenerInstalled` sentinel stored on the registry's `ctx()`. The sentinel dies with the registry, so a new `SVGDocument` whose registry gets allocated at a reused address correctly misses the sentinel and connects fresh.

**P2 — pool eviction**: added `kMaxPoolEntriesPerKey = 8` per-bucket cap (overflow releases drop the texture) and age-out at `beginFrame` (buckets unused for `kBucketEvictAfterFrames = 120` frames ≈ 2s @ 60fps are erased). Steady-state is still a no-op; resize-heavy sessions now cap memory at recently-seen sizes.

Regression sweep green locally:
- `//donner/svg/renderer/geode:geode_perf_tests`
- `//donner/svg/renderer/tests:renderer_geode_golden_tests`
- `//donner/svg/renderer/tests:renderer_public_api_tests` (catches the pre-fix StrokeDasharray crash)
- `//donner/svg/renderer/tests:resvg_test_suite_geode`

PR #564, opened 2026-04-20
CI: adopt bazel-diff target determinator (doc 0031 M2.7)

jwmcglynn:
🤖 ## Summary
- add a `determine-targets` job to `main.yml` that computes affected Bazel targets with bazel-diff on PRs
- fall back to `//...` for infra-touching PRs, non-PR events, and a manual `ci:full-test` label override
- rewire the Linux/macOS Bazel jobs to consume the determinator output and mark doc 0031 M2.7 done

## Notes
- this PR edits `.github/workflows/main.yml`, so the determinator should intentionally report `fallback=true` here
- local YAML/diff sanity checks passed; local `bazel test //...` was attempted twice but the shared arm64 environment's Bazel server crashed before reaching workspace execution


jwmcglynn:
🤖 Landed directly on `feature/ci-hardening-2026q2` as commit 89d41d00 (conflict with M2.8's checkbox in doc 0031 resolved by marking both M2.7 and M2.8 as complete).

PR #592, opened 2026-05-11
Compositor: fix #582 fast-path drag + dragged-shape tracking

### LINKED ISSUE #582: Drag selection shears multi-element composites (<g filter> siblings separate on drag)
## Summary

Dragging a `<g filter>` whose siblings form a single visual composite (e.g. the splash's small lightning bolt — `Lightning_glow_dark` + `Lightning_glow_bright` + `Lightning_white` under a common parent) only moves the clicked `<g filter>`; the siblings stay at the pre-drag position. The composite shears into two visible pieces, reading to users as "the filter disappeared". Same mechanism makes the DONNER text word split on letter drag in `filter_elm_disappear-4.rnr` — each letter is its own `<path>`.

## Repro

Tests in `donner/editor/sandbox/tests/EditorBackendGoldenImage_tests.cc`:

- `EditorBackendCoreFilterDragTest.SingleDragOnLightningGlowDarkKeepsFilter` — minimal: drags `Lightning_glow_dark` once at doc `(474.5, 406.5)→(511.5, 410.5)` (the `-3` recording's drag-2 coords); asserts the post-drag region still has bright-pixel content from the bolt composite. Currently fails with `brightHits = 1763 < 3000` (cold baseline ~4500 at the pre-drag position).
- `EditorBackendCoreFilterDragTest.FilterDisappearRepro3PostSettleMatchesDirectRender` — full `filter_elm_disappear-3.rnr` replay through `EditorBackendCore`. Same failure signature.
- `EditorBackendCoreFilterDragTest.FilterDisappearRepro4DumpFramesForInspection` — diagnostic PNG dumper for `-4.rnr`; `rnr4_after_mup1.png` shows DONNER text split into ``DO`` + ``NNER`` after dragging one letter.

Recordings:
- `donner/editor/tests/filter_elm_disappear-3.rnr`
- `donner/editor/tests/filter_elm_disappear-4.rnr`

## Failed hypotheses investigated

1. **Promoted-layer rasterize skipping the filter** — instrumented `CompositorController::rasterizeLayer` with pixel-color counting. `Lightning_glow_dark`'s bitmap has 19 304 non-zero pixels spread across a 145×178 region, including the expected blurred dark-blue and some bright edge pixels. Gaussian-blurring a `fill=#224084` path naturally produces many pixels near `#224084` — the earlier \"flat-fill\" heuristic was a false positive. Also tried refusing the interaction-promote when the entity is mandatory-promoted via `MandatoryHintDetector`; zero effect on output.
2. **Segment carving leaking raw paths into `bg`** — reviewed `rasterizeDirtyStaticSegments` and `recomposeSplitBitmaps`. Segments exclude promoted layer ranges via paint-order indices; `bg` composes from segments + earlier promoted layer bitmaps (which include filter output), not raw subtree geometry.

## Actual root cause

`SelectTool::ElevateToCompositingGroupAncestor` walks up to the *outermost* `<g filter>` / `<g mask>` / `<g clip-path>` ancestor and stops there. On the splash:

```xml
<g class=\"cls-94\">
  ...
  <g id=\"Lightning_glow_dark\" filter=\"url(#lightning_glow_dark_blur)\">  <!-- dark halo -->
  <g id=\"Lightning_glow_bright\" filter=\"url(#lightning_glow_bright_blur)\">  <!-- cyan inner glow -->
  <g id=\"Lightning_white\">  <!-- bright core, NO filter -->
  ...
  <g id=\"Big_lightning_glow\" filter=\"url(#big_lightning_glow_blur)\">  <!-- separate larger bolt -->
  ...
</g>
```

A click on `Lightning_glow_dark` elevates to `Lightning_glow_dark` (stops there). Dragging moves only it. The three siblings under `cls-94` that form the small bolt don't move together.

## Attempted fix that didn't work

Naive escalation `HasMultipleCompositingChildren(parent)` — if the plain-`<g>` parent has ≥2 children with `filter`/`mask`/`clip-path`, elevate past the filter-g to the parent. Too aggressive: the splash's `<g class=\"cls-94\">` contains `Lightning_glow_dark`, `Lightning_glow_bright`, AND `Big_lightning_glow` (geometrically separate), so the naive fix promotes the entire scene. Reverted.

## Sketch for a proper fix

Elevation needs spatial-proximity awareness. Candidate heuristic: escalate to the parent only when the filter-g's bbox overlaps with another filter/mask/clip-path sibling's bbox at that parent level. Cheap enough to compute (per-click work on a handful of siblings), distinguishes composites from scattered filter uses.

Alternative: stop auto-elevating past the leaf hit at all — require users to explicitly select the parent group via some UI gesture (shift-click to grow selection, or tree navigator). That avoids baking spatial heuristics into the selection code but changes the \"click a filtered element, drag it\" ergonomics that already work for single-filter cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jwmcglynn:
## Summary

First of four Tier-1..4 PRs splitting general-purpose improvements out of the sandbox branch onto `main` (per [design 0032](docs/design_docs/0032-sandbox_branch_split.md)). This PR is the **compositor `#582` bundle** — four cherry-picks from sandbox that restore the `<g filter>` drag fast path and fix three positioning bugs that flowed from the same area.

- **`<g filter>` subtree fast path** (`4048a090`): relax the per-frame eligibility gate so subtree-rooted promoted layers qualify. Drags on `donner_splash.svg` go from ~25 ms/frame back to sub-frame.
- **#582 fix — fast-path subtree translation accumulates across frames** (`f32be87c`): pass the per-frame delta (not cumulative-since-stamp) into `propagateFastPathTranslationToSubtree`; descendants no longer drift forward by `n * frame_delta` after `n` frames.
- **Compose-offset order for resized elements** (`fd60ed9d`): compute canvas-from-canvas directly as `bitmapStamp.inverse() * worldFromEntity` so a resized-then-dragged shape tracks the cursor 1:1 instead of lagging by `oldScale`.
- **`setBitmap` resets `canvasFromBitmap_`** (`48ec5927`): clear the layer's compose offset when the bitmap is restamped, eliminating first-frame jitter after a re-rasterize.

Plus the design-doc work that's part of this bundle:

- `docs/design_docs/0025-composited_rendering.md` gets a ~200-line postmortem of the #582 investigation (root cause, why the dual-path `verifyPixelIdentity` assertion missed it, the four rounds of speculative fixes, what finally worked, follow-up action items).
- `CLAUDE.md` + `AGENTS.md` policy updates distilled from that postmortem: red→green transition required for fix commits, no boutique pixel-diff helpers, no dead/parallel reimplementations, etc.
- `docs/design_docs/0032-sandbox_branch_split.md` is the four-tier split plan itself; lives at the bottom of the editor-dev stack.

Automated coverage in `donner/svg/compositor/CompositorGolden_tests.cc` (`SplashDrag*`, `DraggingFilterGroupSubtreeEngagesFastPath`, `LayerComposeOffsetReflectsDocumentDeltaForResizedElement`, `ScaledCanvasTranslationOnlyDragProducesCorrectPixels`) pins each fix; `bazel test //...` green locally.

## Test plan

- [x] `bazel test //...` green locally.
- [x] Manual smoke on `donner_splash.svg`:
  - [x] Drag `Big_lightning_glow` — smooth, no frame stalls.
  - [x] Long circular drag on a `<g filter>` — cursor and shape stay coincident; no drift.
  - [x] Zoom canvas to non-100%, drag a filter group — shape tracks 1:1, no scale lag.
  - [x] Drag-release-drag — no first-frame jitter on the second drag.
  - [x] Non-filter drags (DONNER text, cloud paths) unaffected.

🤖 PR drafted by Claude (Opus 4.7) per the design 0032 split plan; review and edit as needed.

### DIFF
--- a/donner/svg/compositor/CompositorController.cc
+++ b/donner/svg/compositor/CompositorController.cc
@@ -742,6 +742,26 @@ void CompositorController::renderFrame(const RenderViewport& viewport) {
       unhandledDirtyEntities.reserve(resolutions.size());
       for (const auto& res : resolutions) {
         auto& instance = registry.get<components::RenderingInstanceComponent>(res.entity);
+        // Compute the per-frame delta BEFORE we overwrite the RIC:
+        // `instance.worldFromEntityTransform` at this moment is the
+        // PREVIOUS frame's value (set either by a prior fast-path
+        // iteration, by `prepareDocumentForRendering`, or by the
+        // initial tree instantiation). `res.delta` below, in contrast,
+        // is the cumulative delta since the bitmap was STAMPED —
+        // the right value for `setCanvasFromBitmap` (which operates
+        // against the stamped bitmap) but the WRONG value for
+        // `propagateFastPathTranslationToSubtree`, which left-composes
+        // it onto each descendant's already-translated
+        // `worldFromEntityTransform`. Using the stamp-delta there
+        // re-applies every prior frame's translation on top of
+        // itself, producing canvas pixels that drift further from
+        // the DOM's true position with every drag frame — the #582
+        // "filter element mispositioned after many drag frames"
+        // symptom. Per-frame delta is what the descendants actually
+        // need: layered over their current (already-post-previous-frame)
+        // transforms, it advances them exactly one drag step.
+        const Transform2d perFrameDelta =
+            res.newWorldFromEntity * instance.worldFromEntityTransform.inverse();
         instance.worldFromEntityTransform = res.newWorldFromEntity;
 
         if (res.delta.isTranslation()) {
@@ -754,7 +774,7 @@ void CompositorController::renderFrame(const RenderViewport& viewport) {
           // rather than going back to the renderer.
           res.layer->setCanvasFromBitmap(res.delta);
           if (res.isSubtree) {
-            propagateFastPathTranslationToSubtree(registry, res.entity, res.delta);
+            propagateFastPathTranslationToSubtree(registry, res.entity, perFrameDelta);
           }
         } else {
           // Single-entity layer with non-translation delta (scale,
--- a/donner/svg/compositor/CompositorController.cc
+++ b/donner/svg/compositor/CompositorController.cc
@@ -623,7 +623,11 @@ void CompositorController::renderFrame(const RenderViewport& viewport) {
       Entity entity = entt::null;
       CompositorLayer* layer = nullptr;
       Transform2d newWorldFromEntity;
-      Transform2d delta;
+      /// `bitmapEntity_from_entity` — the canvas-from-canvas mapping from
+      /// the bitmap's stamped entity frame to the entity's CURRENT frame.
+      /// Used as the layer's `canvasFromBitmap` compose offset on the
+      /// bitmap-reuse fast path. **Stamp-relative**, NOT per-frame.
+      Transform2d bitmapEntityFromEntity;
       bool isSubtree = false;
     };
     std::vector<FastPathResolution> resolutions;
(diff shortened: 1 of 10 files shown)


donner/svg/compositor/CompositorController.cc
```diff
@@ -671,8 +671,27 @@ void CompositorController::renderFrame(const RenderViewport& viewport) {
...
+      //
+      // The reverse order (`newWFE * oldEFW`) is an entity-local mapping,
+      // which collapses to the right answer ONLY when the old transform's
+      // linear part is identity. For a previously-resized element it
+      // misreports the canvas delta as `oldScale.inverse() * delta_doc`,
+      // so the cached bitmap moves at `delta_doc / scale` pixels per
+      // drag step while the overlay (driven by worldBounds) moves at
+      // `delta_doc`. See `LayerComposeOffsetReflectsDocumentDeltaForResized
+      // Element` for the regression pin.
+      const Transform2d delta = matchedLayer->bitmapEntityFromWorldTransform()->inverse() *
```
jwmcglynn: Delta needs to be in sourceFromDest notation.  bitmapEntityFromEntity ?

donner/svg/compositor/CompositorController.cc
```diff
@@ -723,6 +742,26 @@ void CompositorController::renderFrame(const RenderViewport& viewport) {
...
+        // it onto each descendant's already-translated
+        // `worldFromEntityTransform`. Using the stamp-delta there
+        // re-applies every prior frame's translation on top of
+        // itself, producing canvas pixels that drift further from
+        // the DOM's true position with every drag frame — the #582
+        // "filter element mispositioned after many drag frames"
+        // symptom. Per-frame delta is what the descendants actually
+        // need: layered over their current (already-post-previous-frame)
+        // transforms, it advances them exactly one drag step.
+        const Transform2d perFrameDelta =
```
jwmcglynn: Same comment as above, Transformd should always be destFromSource. Add AGENTS.md instructions to make this enforced

donner/svg/compositor/CompositorController.cc
jwmcglynn: 🤖 Renamed to `bitmapEntityFromEntity` (both the struct field and the local) in `cf6478df`. The transform takes an entity-current point and returns it in the bitmap-stamp entity frame — `destFromSource` reads correctly. Also strengthened CLAUDE.md + AGENTS.md to require `destFromSource` for every `Transform2d` (locals/fields/params), ban generic names (`delta`, `xform`, etc.), and call out the inverse-rename rule.

donner/svg/compositor/CompositorController.cc
jwmcglynn: 🤖 `perFrameDelta` renamed to `worldFromPreviousWorld` in `cf6478df` (including the `propagateFastPathTranslationToSubtree` parameter and its docstring). It maps a previous-frame world point to its current-frame world position. AGENTS.md §"Transform Naming Convention" now spells out that the rule applies to *every* `Transform2d` and that composition must spell a valid `dest_from_source` chain right-to-left; CLAUDE.md has a concise pointer.

PR #596, opened 2026-05-18
Multithreading support

jwmcglynn:
## Summary

- Add lifetime-aware SVG element handles, guarded document access, and opt-in `ConcurrentDom` support.
- Add detached subtree cleanup, tree mutation helpers, attached-id tracking, and render snapshots that own replay payloads instead of retaining mutable document state.
- Normalize editor and rendering callers around scoped document access, typed mutation helpers, and immutable render snapshots.
- Replace the implementation design doc with shipped developer guides for DOM lifetime and multithreading, plus a public SVG DOM threading/lifetime guide that avoids internal storage details and calls out the more efficient batching APIs.
- Add and update concurrency, lifetime, tree mutation, snapshot, resource-reference, editor async-renderer, and related regression tests.

## Validation

- `bazel test //...`
- `git diff --check HEAD~1..HEAD`
- Earlier targeted checks for this branch included:
  - `bazel test //donner/svg/tests:svg_document_concurrency_tests //donner/svg/tests:svg_tree_mutation_tests //donner/svg/tests:svg_element_lifetime_tests //donner/editor/tests:async_renderer_tests`
  - `bazel test //donner/svg/tests:svg_tests //donner/svg/renderer/tests:renderer_snapshot_tests`
  - `bazel test //donner/editor/tests/...`
  - `bazel build //donner/editor:async_renderer //donner/svg:svg_core`
  - `python3 tools/cmake/gen_cmakelists.py --check`


docs/developer_docs.md
```diff
@@ -28,3 +28,51 @@ Historical and in-flight design docs live under
 [design-docs index](design_docs/README.md) lists every doc with status and a
 one-line summary, and is the right place to start when you want to know *why*
 a subsystem looks the way it does.
+
+## DOM Lifetime And Concurrency
```
jwmcglynn: Move this to its own file. Change to "Multithreading" and add a separate doc for DOM Element lifetime

donner/editor/AsyncRenderer.cc
```diff
@@ -80,6 +80,15 @@ void AsyncRenderer::workerLoop() {
     if (request.renderer != nullptr && request.document != nullptr) {
       ZoneScopedN("AsyncRenderer::workerIteration");
       std::optional<RenderResult::CompositedPreview> compositedPreview;
+      const bool restoreSingleThreadedMode =
```
jwmcglynn: Why is this required? I'd expect that AsyncRenderer unconditionally uses the new infra and is simplified significantly as a result

donner/svg/SVGCircleElement.cc
```diff
@@ -16,53 +16,65 @@ SVGCircleElement SVGCircleElement::CreateOn(EntityHandle handle) {
 }
 
 void SVGCircleElement::setCx(Lengthd value) {
+  DocumentWriteAccess access = handle_.writeAccess();
   invalidate();
 
   auto& properties = handle_.get_or_emplace<components::CircleComponent>().properties;
   properties.cx.set(value, css::Specificity::Override());
+  access.bumpMutationRevision();
```
jwmcglynn: Having to manually bump seems error prone, can we make this implicit on scope exit?

donner/svg/SVGCircleElement.cc
```diff
@@ -16,53 +16,65 @@ SVGCircleElement SVGCircleElement::CreateOn(EntityHandle handle) {
...
+  DocumentWriteAccess access = handle_.writeAccess();
   invalidate();
 
   auto& properties = handle_.get_or_emplace<components::CircleComponent>().properties;
   properties.r.set(value, css::Specificity::Override());
+  access.bumpMutationRevision();
 }
 
 Lengthd SVGCircleElement::cx() const {
+  [[maybe_unused]] DocumentReadAccess access = handle_.readAccess();
```
jwmcglynn, line 46: What's the point of read access? Acquiring a lock? What happens if it isn't acquired?

donner/svg/SVGDocument.cc
```diff
@@ -1,46 +1,110 @@
...
-SVGDocument::SVGDocument(std::shared_ptr<Registry> registry, Settings settings,
+SVGDocument::SVGDocument(SVGDocumentHandle documentState, Settings settings,
                          EntityHandle ontoEntityHandle)
-    : registry_(std::move(registry)) {
-  auto& ctx = registry_->ctx().emplace<components::SVGDocumentContext>(
-      components::SVGDocumentContext::InternalCtorTag{}, registry_);
+    : documentState_(std::move(documentState)) {
+  Registry& registry = documentState_->registry();
+  auto* existingTreeMutations = registry.ctx().find<donner::components::TreeMutationContext>();
+  auto& treeMutations = existingTreeMutations != nullptr
```
jwmcglynn: Is it ever possible to create an SVGDocument without the mutation context? Can we make this unconditional?

Supporting standalone XML use is not a primary scenario, so let's avoid all the complexity of checking for nullptr for the context and std::function - make it set unconditionally

docs/design_docs/0033-dom_lifetime_and_concurrency.md
```diff
@@ -0,0 +1,1074 @@
+# Design: DOM Lifetime and Concurrent Rendering Snapshots
+
+**Status:** Draft
```
jwmcglynn: Convert this to a developer doc, and rename to Multithreading and DOM Lifetime

donner/svg/SVGElement.h
```diff
@@ -26,13 +31,167 @@ class SVGDocument;
...
+
+  /// Test whether the resolved entity has all requested components.
+  template <typename... Type>
+  bool all_of() const {
+    return resolve().all_of<Type...>();
+  }
+
+  /// Get a component from the resolved entity.
+  template <typename Type>
+  decltype(auto) get() const {
```
jwmcglynn: Let's make read/write access less error prone - make it so that if you pass a DocumentWriteAccess it's possible to get a non-const, but if you don't pass it you only get a const.

Apply to all methods, but the ones that mutate are write-access-only (like get_or_emplace)

donner/svg/SVGCircleElement.cc
jwmcglynn: These methods should not need to be modified - only specific hot paths should preacquire read access to avoid churning the lock

jwmcglynn:
🤖 Re: [the P1 review on `donner/svg/SVGElement.cc:268`](https://github.com/jwmcglynn/donner/pull/596#discussion_r3300121077) (self-deadlock in `ElementAnchor::release()`)

Fixed in `9ce82d57` by adding an early-return in `release()` when the calling thread holds **read but not write** access:

```cpp
if (documentHandle_->currentThreadHasAccess() &&
    !documentHandle_->currentThreadHasWriteAccess()) {
  return;
}
```

Write access is reentrant, so the write-held case still proceeds; only the read→write upgrade case deadlocks. The opportunistic detached-node `Collect` we skip when read-held is picked up by the next periodic `NodeLifetimeCollector::Collect` pass (e.g. next source edit, mutation batch commit, end-of-frame), so bailing does not leak.

Regression test added in the same commit: `SVGDocumentConcurrencyTests.ReleaseOfDetachedHandleInsideReadAccessDoesNotSelfDeadlock` — verified red→green locally (TIMEOUT in 20s without the guard, PASS in 0.1s with it).

jwmcglynn:
🤖 Status of your review on this PR — addressed in follow-up #612.

| # | Comment | Status |
|---|---|---|
| [developer_docs.md:32](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262564438) | Move to own file, rename Multithreading, add DOM Element lifetime | ✅ Already addressed in branch commit `6ca78e12` — `docs/multithreading.md` + `docs/dom_element_lifetime.md` exist on main with the matching `\subpage` references |
| [0033-dom_lifetime_and_concurrency.md:3](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262870517) | Convert to developer doc | ✅ Already addressed in `6ca78e12` (the design doc was deleted, dev docs added) |
| [SVGElement.h:142](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262884541) | Type-safe read/write access overloads | ✅ Already in place — `ElementAnchor` has `get<T>() const` / `get<T>(DocumentWriteAccess&) const`, same for `try_get`; `emplace`/`get_or_emplace`/`emplace_or_replace`/`remove` are write-access-only |
| [SVGDocument.cc:23](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262614694) | Make mutation context unconditional, avoid nullptr + std::function checks | 🟡 **Partial in #612** — dropped the redundant per-`std::function` emptiness check at the 5 XMLNode mutation sites; the "always-install-even-for-plain-XML" half is deferred with the editor-wide refactor below |
| [AsyncRenderer.cc:83](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262595246) | Drop `ScopedThreadingModeRestore`, use ConcurrentDom unconditionally | ⏸ **Deferred** — see below |
| [SVGCircleElement.cc:24](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262597354) | Implicit invalidate on scope exit | ⏸ **Deferred** — `invalidate()` is per-element-class (each removes its own `Computed*` components); implicit version needs a polymorphic hook or per-class scoped wrapper |
| [SVGCircleElement.cc:46](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262601608) | What's the point of read access? | ⏸ **Deferred** — see below |
| [SVGCircleElement.cc:46](https://github.com/jwmcglynn/donner/pull/596#discussion_r3262886918) | Only hot paths should preacquire; revert generated readAccess | ⏸ **Deferred** — see below |

**Why the deferred items need an editor-wide refactor:** Comments 2 + 4 + 8 are coupled. Dropping the AsyncRenderer flip-back (so the document stays in `ConcurrentDom`) is what makes the per-getter `readAccess()` reverts safe, but it also requires the editor's UI frame loop to hold coarse-grained read access — otherwise UI-thread reads in `runFrame`, the sandbox harness, and the test fixtures all Class-B-assert on the unguarded reads. I verified this by trying the AsyncRenderer simplification in isolation: `async_renderer_tests` and `structured_editing_stress_tests` immediately hit `SVGElement.cc:253` asserts, so I reverted that change in #612 rather than ship a broken intermediate.

The full refactor (drop the flip → plumb coarse-grained read access through the editor frame loop and harnesses → revert ~147 per-getter `readAccess()` calls → design the per-class implicit-invalidate hook) is a multi-session effort I'd rather plan deliberately. Happy to take that as a separate tracked follow-up.

If any of the "already addressed by `6ca78e12`" entries don't match what you want, just say the word and I'll iterate.

Issue #542, opened 2026-04-18
Intel Arc A380 + Mesa ANV Vulkan: flaky GPU hang in RendererGeode tests (cumulative-submission fence-signal race)


jwmcglynn:
## Summary

On Intel Arc A380 (DG2 / vendorID 0x8086, deviceID 0x56a5) running Mesa ANV Vulkan on the Linux Xe KMD kernel driver, `RendererGeode` exhibits a non-deterministic GPU hang on `device.poll(true)` during `takeSnapshot()`. Single tests in isolation mostly pass; cumulative submission in longer test sequences accumulates probability until the run consistently hangs. The hang is not in any Geode-specific shader — even empty render passes (clear + texture-to-buffer copy, no draws) hit it.

This issue is a write-up so we have a durable repro + context to hand to Mesa ANV / Xe-KMD maintainers when we eventually file upstream. Not for upstream submission right now.

## Environment

- **Host**: aarch64 Ubuntu 24.04 (Neoverse-N1 + PCIe Intel Arc A380)
- **Kernel**: 6.17.0-20-generic
- **KMD**: Xe (`lsmod | grep xe` = 3.7MB module)
- **Vulkan ICD**: Mesa ANV (`Intel open-source Mesa driver`)
- **Mesa versions tested**:
  - 25.2.8-0ubuntu0.24.04.1 (Ubuntu stock) — baseline hang rate
  - 26.0.5~kisak1~n (kisak-mesa PPA) — reduced but not eliminated
- **WebGPU/wgpu-native**: bundled in `//third_party/wgpu-native`, see `donner/svg/renderer/geode/GeodeDevice.cc`
- **GPU**: PCIe Intel Arc A380 Graphics (DG2), `driverID = DRIVER_ID_INTEL_OPEN_SOURCE_MESA`

## Repro (post-PR #540 on main)

```
bazel test --config=geode //donner/svg/renderer/tests:renderer_geode_tests \
  --test_output=errors --cache_test_results=no --test_timeout=180
```

On Arc, this consistently hangs at `PushPopTransform` (6th test in the sequence) or later. Force-timeout at 180s.

llvmpipe reproduces nothing — the issue is Arc-specific:

```
VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/lvp_icd.json \
bazel test --config=geode --test_env=VK_ICD_FILENAMES \
  //donner/svg/renderer/tests:renderer_geode_tests
# always passes
```

[...]

PR #615, opened 2026-05-27
editor: drop AsyncRenderer threading-mode flip-back (#596 follow-up)

jwmcglynn:
Follow-up to PR #596, addressing the AsyncRenderer threading-mode flip-back review feedback and keeping the public DOM API within the existing ECS escape-hatch boundary.

## What this PR does

**AsyncRenderer simplification.** The worker no longer flips the document between `SingleThreaded` and `ConcurrentDom` on every render iteration. On first render it transitions the editor document into `ThreadingMode::ConcurrentDom` and leaves it there for the editor lifetime. UI-thread reads are responsible for explicit guards (`withReadAccess` / scoped `DocumentReadAccess`) where they touch the live document.

**Existing ECS escape hatch only.** The proposed public `SVGElement::entity()` accessor was removed. Call sites that only need the entity id now use the existing documented advanced escape hatch: `unsafeEntityHandle().entity()`.

**Guarded mutation entry points.** All `SVGDocument` mutation entry points that resolve `SVGElement` handles now acquire `writeAccess()` up front: `applySourceEdit`, `setElementAttribute`, `removeElementAttribute`, `insertElement`, and `removeElement`.

**Coverage CI robustness.** `tools/coverage.sh` now uses an OS-aware file mtime helper so Linux does not accidentally parse GNU `stat -f` filesystem metadata as an integer timestamp when checking for stale coverage reports.

## Still Deferred

Comments 4 / 8 / 3 from PR #596's review remain separate follow-ups: removing the per-getter `DocumentReadAccess` lines in SVG element classes, and adding an implicit invalidate-on-scope-exit hook.

## Validation

Local checks for this follow-up:

```
bash -n tools/coverage.sh
git diff --check
```

Prior branch validation before the coverage-script follow-up:

```
bazel test //donner/editor/... //donner/svg/... //donner/base/... //tools/...
```

PR CI is running on the latest push.

### DIFF
--- a/donner/editor/AsyncRenderer.cc
+++ b/donner/editor/AsyncRenderer.cc
@@ -49,23 +49,6 @@ RenderResult::CompositedPreview BuildFullCanvasCompositedPreview(
   };
 }
 
-/// Restores a document's threading mode to SingleThreaded when a worker iteration exits.
-///
-/// The async worker flips the (UI-owned, normally SingleThreaded) document into ConcurrentDom for
-/// the duration of a render so it can hold an explicit \ref donner::svg::DocumentWriteAccess guard.
-/// This restores the prior mode on every exit path. Declare it *before* the write-access optional
-/// so the write access is released first.
-struct ScopedThreadingModeRestore {
-  svg::SVGDocument* document = nullptr;
-  bool restore = false;
-
-  ~ScopedThreadingModeRestore() {
-    if (restore && document != nullptr) {
-      document->setThreadingMode(svg::ThreadingMode::SingleThreaded);
-    }
-  }
-};
-
 }  // namespace
 
 PresentationSnapshotPlan ChoosePresentationSnapshotPlan(bool hasCompositedPreview,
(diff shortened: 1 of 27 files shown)


jwmcglynn:
🤖 Addressed both review comments in `3e2cccfd` (force-pushed).

**[P1: Guard all public mutation entry points](https://github.com/jwmcglynn/donner/pull/615#discussion_r3308019330):**
Right — `setElementAttribute` and `removeElement` weren't the full set. All `SVGDocument` mutation entry points that resolve an `EntityHandle` from an `SVGElement&` now acquire `writeAccess()` up front: `setElementAttribute`, `removeElementAttribute`, `insertElement`, `removeElement`, `applySourceEdit`. That covers the `DocumentSyncController` writeback and `AsyncSVGDocument` insert paths you flagged (they route through these entry points).

**[P2: Avoid adding an ECS entity to the public DOM API](https://github.com/jwmcglynn/donner/pull/615#discussion_r3308019332):**
Fair call against `AGENTS.md`'s Public API Boundary. Reverted the `SVGElement::entity()` accessor entirely. The ~115 `.entityHandle().entity()` call sites now use `.unsafeEntityHandle().entity()` — the existing documented escape-hatch path. No new ECS-typed methods on the public DOM facade.

Diff: 26 files changed, 342 insertions / 344 deletions (net **-2 lines**). Verbose at call sites but uses only pre-existing API.

`bazel test //donner/editor/... //donner/svg/... //donner/base/... //tools/...` → **612 pass, 6 fail (all environmental Geode #542), 11 skipped — zero non-Geode failures.**

Issue #625, opened 2026-05-29
Geode: feGaussianBlur/complex-transform analytic-vs-finite-sample edge on rotated blur


jwmcglynn:
🤖 AI-filed (Claude). The only remaining Geode-disabled resvg test after the analytic dual-ray Slug coverage work (docs/design_docs/0041 §6).

**Test:** `filters/feGaussianBlur/complex-transform.svg` (Geode backend only; CPU passes).

**Symptom:** ~259px diff vs the resvg reference — a ~1px edge band on the rotated/skewed rect edges, amplified by the blur kernel. Geode's analytic Slug coverage produces a finer/correct edge that differs from resvg's finite-sample scan-converter; the blur spreads the sub-pixel difference past the strict budget.

**Status:** Geode content is structurally correct (this is the analytic-vs-finite-sample divergence 0041 §6 documents, not a bug in Geode). Currently gated via `disableBackend(Geode, ...)`.

**Resolution options:** (a) a per-backend Geode golden (the pattern used for `feColorMatrix non-normalized` and `feImage/svg`, where Geode is verified-correct) — preferred if Geode's blurred output is confirmed correct by inspection; or (b) keep gated as a documented analytic-vs-finite-sample case.

Issue #624, opened 2026-05-29
tspan/textPath positioning: paint-order on-tspan, transform-origin on-text-path


jwmcglynn:
🤖 AI-filed (Claude) while triaging resvg-test-suite skips. See docs/design_docs/0021-resvg_feature_gaps.md.

Two skipped cases that share a tspan/textPath positioning root (NOT the feature they're filed under):

- **`painting/paint-order/on-tspan`**: residual is tspan-split glyph positioning drift, present even in the default-order run — not a paint-order bug.
- **`structure/transform-origin/on-text-path`**: transform-origin pivot not composed with textPath content placement.

Issue #623, opened 2026-05-29
Marker & stroke-dasharray edge cases (resvg suite residuals)


jwmcglynn:
🤖 AI-filed (Claude) while triaging resvg-test-suite skips. See docs/design_docs/0021-resvg_feature_gaps.md.

Small residual rendering diffs left skipped after the paint-order / multiple-closepath / `0 N` dasharray fixes:

- **`painting/stroke-dasharray/n-0`** (`40 0` pattern): ~625px closed-path dash-start-vertex seam.
- **`painting/marker/marker-on-rounded-rect`**: an extra start marker is drawn (~666px).
- **`painting/marker/recursive-5`**: ~787px diff (recursive marker reference).

Each is a narrow geometry/vertex-set difference; root-cause individually.

jwmcglynn:
Resolved by the merged dasharray and marker fixes: #864 corrected the closed-path `N 0` seam, #867 corrected marker vertex/recursion handling, and #858 removed the obsolete rounded-marker override once the residual fit the suite default. Verified again on current main `6d86196d7`: `n-0.svg` passes exact identity on TinySkia and Geode; `marker-on-rounded-rect.svg` passes at 24px/18px respectively under the standard 100px budget; and `recursive-5.svg` passes at 0px/8px. All three original acceptance cases are active, not skipped.

Issue #622, opened 2026-05-29
<a> hyperlink element: text children dropped (no SVGAElement)


jwmcglynn:
🤖 AI-filed (Claude) while triaging resvg-test-suite skips. See docs/design_docs/0021-resvg_feature_gaps.md (smaller feature gaps).

**Gap:** `<a>` parses as `SVGUnknownElement`, which lacks the `TextComponent`/`TextPositioningComponent` the text layout descends into, so an `<a>`'s text children are dropped (blank render). Needs a real `SVGAElement` acting as a text-content grouping/container element (like `<g>`/`<tspan>`), not just a renderer tweak.

**Affected tests (currently skipped):** `structure/a` (3).

PR #626, opened 2026-05-29
resvg suite: Geode analytic Slug AA + close CPU/Geode feature gaps

jwmcglynn:
Two major thrusts on the resvg-test-suite branch:

### 1. Geode: analytic Slug dual-ray anti-aliasing (replaces 4× MSAA)
Geode now computes **official Slug analytic dual-ray coverage at 1 sample/pixel on every adapter**, replacing the 4× MSAA `sample_mask` path and the Intel-Arc alpha-coverage fallback. The Mac (Metal MSAA) vs Linux (Intel/Vulkan) divergence is gone, and `GeodeTinyParity` is retired (tiny-skia is a finite-sample scan-converter; each backend now gates against the shared reference via `GeodeGolden`/`TinyGolden`). Design + verbatim reference algorithm: `docs/design_docs/0041`.
- Encoder: X-monotonic vertical bands + dense band-grid lookup + single bounding quad (seam-free folded coverage).
- Shaders: `slug_fill`/`gradient`/`mask` rewritten to the reference `CalcCoverage` combine; `*_alpha_coverage.wgsl` + MSAA plumbing deleted.

### 2. Closed a batch of resvg feature gaps (red→green, un-skipped)
**CPU backend:** feImage embedded/subregion/PAR (bicubic), feTile empty-region, feDiffuseLighting no-light; F7 `paint-order` (shapes+text), F12 `transform-origin` on text/image, `<symbol>` transform, stroke-dasharray `0 N` caps, nested-svg `overflow`, marker multiple-closepaths.

**Geode (the 16):** these were **misdiagnosed** as "slug_fill edge-coverage" — the analytic rewrite left them byte-identical, proving coverage wasn't the cause. They were three real bugs (pattern-tile filter-region-scissor leak → all 10 feConvolveMatrix + feMorphology; feMorphology linearRGB round-trip; degenerate zero-area closed-stroke decomposition → marker) plus two verified-correct cases given per-backend Geode goldens (feColorMatrix non-normalized, feImage/svg). All 16 un-skipped and green.

## Validation
- `bazel test //...` → **622 pass, 0 fail**, 57 skipped (all variants incl. `*_geode`/`*_text_full`/`*_tiny`).
- `python3 tools/cmake/gen_cmakelists.py --check` → passed.
- clang-format 18 → clean.

## Remaining / deferred (tracked)
#621 (transform-origin on paint-servers), #622 (`<a>` needs SVGAElement), #623 (marker/dash residuals), #624 (tspan/textPath positioning), #625 (feGaussianBlur/complex-transform analytic-vs-finite-sample).


jwmcglynn:
🤖 Thanks @codex — both P2s addressed in 45681126:

1. **Cubic X-monotonic split** (`Path.cc`): the cubic branch now uses the axis-selected `cubicExtrema` lambda instead of `CubicYExtrema`, so `toMonotonic(MonotonicAxis::X)` honors the axis for cubic-containing paths. Added regression test `Path.ToMonotonicXSplitsCubicAtXExtrema` (verified red→green: fails on the old `CubicYExtrema`, passes with the fix).
2. **Geode paint-order component switches** (`RendererGeode::drawPath`): now honors `PaintParams::drawFillComponent`/`drawStrokeComponent`, so non-canonical `paint-order` reorders rather than double-paints on Geode. Both flags default to true → no-op for ordinary fill-then-stroke draws (existing tests unchanged); `drawRect`/`drawEllipse` delegate to `drawPath` so they're covered.

Issue #603, opened 2026-05-23
Audit and constrain non-owning Ref/view return types (dangling-reference footgun)


jwmcglynn:
🤖 ## Background

During the multithreading rebase (PR #596), `SVGElement::tagName()` returned an `xml::XMLQualifiedNameRef` (a non-owning view type). Editor code that cached `tagName().name` as a `std::string_view` across statements dereferenced freed memory once the temporary `Ref` was destroyed — an ASAN `stack-use-after-scope` (it surfaced via `SidebarPresenter::BuildTreeNodeLabel`). With `RcString`'s small-string optimization the bytes live inline in the temporary, so the dangling view read freed stack memory.

`tagName()` is being changed to return an owning `xml::XMLQualifiedName` to remove that specific footgun.

## Task

Audit the codebase for other places that **return** (or store as members) `*Ref` view types (`XMLQualifiedNameRef`, `RcStringOrRef`, `std::string_view`, `std::span`, and similar non-owning views) where the referent can be a temporary — these are latent dangling-reference bugs of the same class.

- Enumerate functions/methods that return `*Ref`/view types and classify each: safe (points into stable storage with a documented lifetime contract) vs. risky (can view a temporary).
- Fix or document the risky ones (return owning, or document the lifetime contract clearly).
- Consider a project guideline / lint: **`*Ref` view types are allowed as in-parameters only, not as return types or data members** ("limit Ref to in-values"). Evaluate enforcing via a banned-patterns lint or a clang-tidy check.

## Acceptance
- [ ] Inventory of `*Ref`/view-returning APIs with risk classification.
- [ ] Risky returns converted to owning or given an enforced lifetime contract.
- [ ] Decision on (and, if adopted, enforcement of) the "Ref = in-values only" guideline.

PR #640, opened 2026-05-31
base: constrain non-owning view returns with [[clang::lifetimebound]] (#603)

jwmcglynn:
Refs #603 (focused fix + documented inventory of the rest).

## Root cause
`XMLQualifiedNameRef` (returned by value from `tagName()`) holds `RcStringOrRef` members. Caching `tagName().name` as a `std::string_view` across statements narrows a member of the destroyed temporary Ref to a view; with `RcString`'s small-string optimization those bytes live inline in the temporary, so the cached view reads freed stack memory (the reported ASAN stack-use-after-scope via `SidebarPresenter::BuildTreeNodeLabel`). The leak point is the `std::string_view` conversion on `RcString` / `RcStringOrRef`.

## Change
Annotate every view/pointer accessor that aliases internal storage with `[[clang::lifetimebound]]` on `RcString` and `RcStringOrRef` (`operator string_view`, `data()`, iterators). The reported pattern `std::string_view sv = tagName().name;` now produces a `-Wdangling` diagnostic — a hard error under the clang-tidy gate (`clang-diagnostic-dangling` + `WarningsAsErrors: "*"`).

Did **not** change `tagName()`'s return type (spans the `ElementLike` concept, `FakeElement`, 112 call sites); the `XMLQualifiedNameRef(const XMLQualifiedName&)` constructor is safe (it addrefs, not aliases).

## Tests
- `RcString.ViewLifetimeContract` — pins the safe view contract + ASAN tripwire.
- `XMLQualifiedNameRefTest.OwnsBytesFromQualifiedName` — a Ref built from an owning `XMLQualifiedName` retains its bytes.

Pass normal + `--config=asan`; svg/css/editor/compositor consumers compile clean (no live dangling-bind site exists today).

## Guideline
AGENTS.md now codifies "Ref = in-values only": non-owning views are in-parameters, not returns/members, unless the lifetime contract is documented and enforced with `[[clang::lifetimebound]]`.

## Remaining inventory (follow-up, not in this PR)
`std::span` member-view returns (FontManager, Stylesheet, Path, PropertyRegistry, …), `ChunkedString::firstChunk`, `XMLTokenType::text(source)`, and the `tagName()` family would all benefit from the same annotation; lower-risk than the SSO-temporary dangle fixed here.

---
**Note:** the agent flagged a *pre-existing*, unrelated `misc-include-cleaner` clang-tidy issue in `donner/base/xml/XMLEscape.cc` (missing direct includes) — not touched here; worth a separate narrow fix.

### DIFF
--- a/donner/base/RcString.h
+++ b/donner/base/RcString.h
@@ -111,8 +111,11 @@ public:
     return *this;
   }
 
-  /// Cast operator to `std::string_view`.
-  operator std::string_view() const {
+  /// Cast operator to `std::string_view`. With the small-string optimization the returned view can
+  /// point into this object's inline buffer, so it must not outlive the \ref RcString. Annotated
+  /// `[[clang::lifetimebound]]` so Clang's `-Wdangling` and clang-tidy flag binding the view to a
+  /// temporary \ref RcString (the small-string-in-temporary dangle from issue #603).
+  operator std::string_view() const [[clang::lifetimebound]] {
     return data_.isLong() ? data_.long_.view() : data_.short_.view();
   }
 
@@ -215,9 +218,13 @@ public:
   /// @}
 
   /**
-   * @return a pointer to the string data.
+   * @return a pointer to the string data. With the small-string optimization the pointer can alias
+   *   this object's inline buffer, so it must not outlive the \ref RcString (annotated
+   *   `[[clang::lifetimebound]]`).
    */
-  const char* data() const { return data_.isLong() ? data_.long_.data : &data_.short_.data[0]; }
+  const char* data() const [[clang::lifetimebound]] {
+    return data_.isLong() ? data_.long_.data : &data_.short_.data[0];
+  }
 
   /**
    * @return if the string is empty.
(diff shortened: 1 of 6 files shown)


donner/base/RcString.h
```diff
@@ -215,9 +218,13 @@ class RcString {
   /// @}
 
   /**
-   * @return a pointer to the string data.
+   * @return a pointer to the string data. With the small-string optimization the pointer can alias
+   *   this object's inline buffer, so it must not outlive the \ref RcString (annotated
+   *   `[[clang::lifetimebound]]`).
    */
-  const char* data() const { return data_.isLong() ? data_.long_.data : &data_.short_.data[0]; }
+  const char* data() const [[clang::lifetimebound]] {
```
jwmcglynn: We still need to run in non-clang environments.  Add a helper to base/Utils.h and also have a gcc version if possible

jwmcglynn: Fixed in 0758421e7. The raw [[clang::lifetimebound]] annotations are now wrapped in UTILS_LIFETIME_BOUND in donner/base/Utils.h, and RcString/RcStringOrRef use that macro. It maps to [[clang::lifetimebound]] on Clang, [[msvc::lifetimebound]] on MSVC, and expands to nothing on GCC because there is no equivalent GCC lifetime-bound attribute available.

PR #658, opened 2026-06-30
Fix main CI failures; use RE for self-hosted Linux CI

jwmcglynn:
Fixes the current main-branch CI failures without changing the performance thresholds.

- Add the missing Bazel `--config=ubsan` configuration used by the sanitizer workflow.
- Fix an ASan-reported dangling `std::string_view` in `StyleSystem::updateStyle` by retaining the optional `RcString` backing storage for the parse.
- Bound recursive path boolean contour search so degenerate inputs return `TooComplex` instead of timing out, and add the timeout-triggering fuzzer corpus seed.
- Update async renderer wallclock tests to mutate `AsyncSVGDocument` DOM elements through `withWriteAccess()` when simulating transforms.
- Remove Linux self-hosted workflow fallback flags so the runner-provided `--config=ci` remote execution policy is not overridden.

RE changes:
- Remove Linux self-hosted fallback Bazel flags from CI and Coverage workflows.
- Stop overriding the runner-provided `--config=ci` remote execution policy with `--remote_local_fallback=true`.
- Run self-hosted Linux CI fetch/build/test with `--config=latest_llvm` so RE actions use Donner's hermetic LLVM toolchain instead of runner-local Nix clang paths.
- Keep the macOS self-hosted fallback behavior unchanged.


donner/svg/renderer/geode/GeodeDevice.cc
```diff
@@ -32,6 +34,89 @@ void OnUncapturedError(WGPUDevice const* /*device*/, WGPUErrorType type, WGPUStr
                message.data ? message.data : "");
 }
 
+bool EqualsAsciiCaseInsensitive(std::string_view a, std::string_view b) {
```
jwmcglynn, line 37: Use StringUtils instead

PR #673, opened 2026-07-02
coverage: cache test results; capture coverage diagnostics

jwmcglynn:
Cuts coverage wall time by removing the biggest structural cost: `test:coverage --nocache_test_results` forced **every** coverage run to re-execute **every** test in scope, even unchanged ones (and instrumented tests run slower than normal ones).

## Why caching is sound

A cache-hit test's `coverage.dat` is part of its cached result; `--experimental_fetch_all_coverage_outputs` makes Bazel download coverage outputs for cached tests so the combined lcov report stays complete. Validated locally: two consecutive coverage runs of the same targets, second run fully cached (`Executed 0 out of 2 tests`), **byte-identical** `_coverage_report.dat` (same md5, all `SF:` records present).

The flag dates to 2024-04 ("Fix code coverage on mac"), before the remote cache and affected-target selection existed — re-running everything was the safe default then; now it's the dominant coverage cost. With caching, PR coverage re-runs only tests whose actions actually changed.

## Also in this PR

- **Coverage-lane diagnostics**: `tools/coverage.sh` honors `DONNER_CI_DIAGNOSTICS_DIR` (inner `bazel coverage` profile + BEP, phase wall-times for bazel/filter/genhtml), uploaded by `coverage.yml` as `ci-diagnostics-coverage-self-hosted-*` (3-day retention) — same artifact contract as the CI lane.
- **No `bazelisk shutdown` between Fetch and Generate coverage** (reuses the warm server's analysis, same fix as the CI lane).

## Expected effect

Full-fallback coverage baseline was ~8m11s of `Generate coverage`. With warm cache, an ordinary PR's coverage should re-run only affected tests; unchanged-heavy PRs drop to instrumented-compile + report-merge time. This PR's own coverage run is the first measurement (it falls back to full scope since `.bazelrc` changed — worst case, but with cold coverage cache so treat it as the seeding run; the next PR gets the real number).

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -228,7 +228,14 @@ build:coverage --action_env=BAZEL_LLVM_COV=llvm-cov
 build:coverage --action_env=BAZEL_LLVM_PROFDATA=llvm-profdata
 
 test:coverage --collect_code_coverage
-test:coverage --nocache_test_results
+# Coverage test results ARE cacheable: a cache-hit test's coverage.dat is part
+# of its cached result, and --experimental_fetch_all_coverage_outputs makes
+# Bazel download it so the combined lcov report stays complete (validated
+# 2026-07-02: two consecutive runs, second fully cached, byte-identical
+# _coverage_report.dat). The previous `--nocache_test_results` (2024-04,
+# pre-remote-cache) forced every coverage run to re-execute every test in
+# scope, which dominated CI coverage wall time.
+test:coverage --experimental_fetch_all_coverage_outputs
 
 # Basic ASAN/UBSAN that works for gcc
 build:asan --config=latest_llvm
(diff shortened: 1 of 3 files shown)


jwmcglynn:
🤖 Fixed the hosted-lane failures: cacheable coverage requires `--experimental_split_coverage_postprocessing` — without it the test spawn's `_coverage/gcov` is an absolute symlink into the ephemeral sandbox and cache-upload validation rejects it ("child gcov is a dangling symbolic link"). `--nocache_test_results` had been masking exactly this. Validated locally: fresh execution passes with a byte-identical combined report. This is Bazel's documented flag pairing for cached/remote coverage.

jwmcglynn:
🤖 Also fixed the burst-failure mode seen across tonight's coverage runs: the self-hosted lane's `--remote_timeout=5s` (a pre-RE fail-fast relic for the old LAN cache) made large coverage uploads fail with `DEADLINE_EXCEEDED` whenever the RE endpoint was busy. Raised to 60s — dead-endpoint failures still surface immediately via connection errors, and `--remote_local_fallback` still covers outages. (Server-side upload throughput was also improved in parallel.)

PR #674, opened 2026-07-02
test: use TEST_TMPDIR for editor-control rnr scratch files

jwmcglynn:
`EditorControlSession_tests` wrote **fixed filenames** into `std::filesystem::temp_directory_path()`, which on remote-execution workers (no `TMPDIR` set) resolves to the shared `/tmp`. A file left there by one worker user makes later runs under a different user fail to open the same path (sticky bit): `RecordsAndReplaysRnrFromSelectorDrag` failed on the macOS self-hosted lane with `svg.is_open()==false` — `/tmp/donner_editor_control_rnr_roundtrip.svg` was owned by another user.

Fix: prefer `TEST_TMPDIR` (Bazel's per-test scratch, always writable) with a `temp_directory_path()` fallback for non-bazel runs.

**Follow-up (not this PR):** the same pattern exists in `DocumentSave_tests`, `RnrReplay_tests`, `ImageComparisonTestFixture`, and `RenderPanePresenterVisualRepro_tests` — worth a shared `TestTempDir()` helper in `base_test_utils` and a sweep.

**Operator note:** the poisoned files on the shared macOS host need a one-time cleanup (`sudo rm -f /tmp/donner_editor_control_rnr_roundtrip.*`) before reruns of affected jobs go green.

### DIFF
--- a/donner/base/BUILD.bazel
+++ b/donner/base/BUILD.bazel
@@ -76,8 +76,11 @@ donner_cc_library(
         "tests/BaseTestUtils.h",
         "tests/ParseResultTestUtils.h",
         "tests/Runfiles.h",
+        "tests/TestTempDir.h",
     ],
-    visibility = donner_internal_visibility(),
+    # //tools tests (MCP servers, build tooling) share the same test-hygiene
+    # helpers (TestTempDir) as //donner tests.
+    visibility = donner_internal_visibility() + ["//tools:__subpackages__"],
     deps = [
         "@com_google_gtest//:gtest",
         "@rules_cc//cc/runfiles",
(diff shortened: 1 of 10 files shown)


jwmcglynn:
🤖 Addressed: dropped the host/date-specific parenthetical from the `TestTempDir` comment; the rationale is now fully self-contained. (Nice to see the new no-private-infra rule enforced within hours of landing.)

jwmcglynn:
🤖 Extended with the full sweep: shared `donner::TestTempDir()` helper in `base_test_utils` (visibility widened to `//tools`), applied to DocumentSave/RnrReplay/ImageComparisonTestFixture/RenderPanePresenterVisualRepro + the session test. All five targets pass locally. This closes the whole remote-worker shared-temp collision class, including the `render_pane_presenter_visual_repro_tests` and `text_editor_tests` remote failures seen earlier.

PR #680, opened 2026-07-02
ci: cache bazel-diff base hashes keyed by base SHA

jwmcglynn:
`determine-targets` runs `bazel-diff generate-hashes` twice per PR (base + head) — roughly half of its 45–60s wall time, all of it serial before any self-hosted runner starts. Base hashes are a pure function of (bazel-diff version, base SHA), so:

- **PR runs** restore `~/.cache/bazel-diff-base-hashes/<base_sha>.json` via `actions/cache` and skip the base worktree + hash generation entirely on a hit; on a miss they generate and save (warming subsequent pushes to the same PR).
- **Main pushes pre-warm** the cache for their own SHA (main-scope caches are visible to all PR branches), so the first PR against any main commit already hits.

Same key convention in `main.yml` and `coverage.yml`, so one pre-warm serves both determinators.

Expected effect: determine-targets drops ~15–25s on every PR run — straight off the serial critical path ahead of the runner jobs. Failure behavior is unchanged: cache miss or unreadable file falls back to generating hashes exactly as today.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -67,6 +67,24 @@ jobs:
         if: github.event_name == 'pull_request'
         run: curl -fsSL -o "$RUNNER_TEMP/bazel-diff.jar" "https://github.com/Tinder/bazel-diff/releases/download/${BAZEL_DIFF_TAG}/bazel-diff_deploy.jar"
 
+      - id: pr-meta
+        name: Extract PR base SHA for hash caching
+        if: github.event_name == 'pull_request'
+        shell: bash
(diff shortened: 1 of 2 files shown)


jwmcglynn:
🤖 Added `-tt Rule` filtering (with `--includeTargetType` hash generation) so source-file labels like `//donner/base:Transform.h` stop reaching the build/test target lists — kills the "is a source file, nothing will be built" warning wall seen on PR #672's run and un-inflates the target counts. Cache key bumped since the hash JSON schema changed.

PR #706, opened 2026-07-03
Text tool: in-canvas editing sessions with caret, box text, and style toggles

jwmcglynn:
Reworks the Text tool from single-click insert into design-tool text sessions, and fixes editor QA issues found while testing (text picking, high-zoom clipping, canvas scrollbars, text selection chrome, and a compositor settle bug that made the first resize of a group pop back).

## Text tool

- **Gestures**: clicking existing text opens an editing session on that element with the caret at the clicked character (trailing-half clicks land after the glyph); dragging a rect creates box text; double-clicking empty canvas starts point text; a plain click on empty canvas creates nothing (it only commits the active session). `MouseModifiers` carries a `doubleClick` bit fed from ImGui's `IsMouseDoubleClicked` on the live-pointer path and press time/position proximity on the document-space replay path.
- **Point and box text**: box text greedy-wraps to the box width using measured character advances, with the box recorded as `data-donner-text-box-width`/`-height`. Soft-wrapped continuation lines carry `data-donner-soft-wrap` so a later editing session reconstructs the logical content losslessly (hard breaks stay hard, wraps recompute).
- **Editing existing text**: session state (content, origin, box, font size) reconstructs from the DOM; chrome maps through the element transform so transformed text gets a correctly placed caret; all DOM reads run inside scoped accesses (ConcurrentDom-safe, regression-tested).
- **Editing**: typing, Enter (hard break), Backspace/Delete, and arrow/Home/End caret movement edit the live `<text>` element through the DOM mutation seam. Multi-line content becomes one `<tspan>` per display line. Cmd+B/I/U toggle `font-weight`/`font-style`/`text-decoration`.
- **Commit semantics**: Escape, clicking away, or switching tools commits the session as one undoable operation — "Insert text" for created elements, "Edit text" for existing ones (skipped when nothing changed), and emptying an existing element deletes it as an undoable "Delete text". An empty created session deletes the element and restores the prior selection.
- **Chrome**: the caret bar and box frame render through `OverlayRenderer` direct to the framebuffer, mirroring the pen hover chrome plumbing.
- New text writes fill through `style="fill: ..."` — the channel the fill-color picker edits — instead of a `fill` presentation attribute.

## Text selection chrome

- The rect around text is its **frame**: the box the user dragged out (`data-donner-text-box-*`) for box text, or the computed ink extent for double-click point-text spans. `SelectionAabb` gains `AuthoredTextBoxLocal` + `TextWorldFrameBounds`, and selection bounds/overlay AABBs anchor to it.
- With the **text tool**, dragging a corner handle resizes the frame only: box attributes rewrite (anchored at the opposite corner) and the content rewraps — glyphs never scale. Resizing point text writes box attributes for the first time, converting the computed frame into a user-sized one. The rotate ring rotates the element; clicks inside the frame but off every glyph park the caret at the end.
- With the **select tool**, handles keep the normal scale/rotate transform semantics.
- Selected text gets the same overlay rectangle and drag/rotate/resize handles as shapes (`SelectionAabb` gains `CollectRenderableTextRoots` + `TextWorldInkBounds`; `SnapshotSelectionWorldBounds` and the occlusion snapshot include text, so SelectTool's handle hit-testing, resize, rotate, and drag-move work on text with no gesture-code changes).
- A baseline underlay renders beneath the selection chrome: one document-space segment per laid-out line, grouped by text-local baseline y from the char-geometry API, in the control-line guidance style.
- The render-pane context menu gains **Convert Text to Outlines** when the clicked element is text (or the selection is all text), delegating to the same DOM-level conversion as the menu bar.

## QA fixes

- **Text picking**: clicking or marqueeing rendered text selects the `<text>` element via glyph ink bounds; hit-test APIs widen from `SVGGeometryElement` to `SVGGraphicsElement`.
- **High-zoom zoom/pan clipping**: `RendererGeode::pushClip` composes clip rects with the current transform, and the direct-present tile drawer left the previous tile's texture transform active — only the first overview-infill region scissored correctly. Fixed by resetting the transform before each infill clip rect.
- **Canvas scrollbars**: the render pane no longer window-scrolls its overlay chrome. New emulated scrollbars (`donner/editor/CanvasScrollbars`) represent the pane's position within the document extent and pan the canvas when dragged; rails are excluded from canvas tool input.
- **First resize of a group popped back** (reported against convert-text-to-outlines output): during a scale/rotate gesture on a promoted subtree layer, the compositor's transform-only fast path reused the cached bitmap and updated the root `RenderingInstanceComponent`, but only pure-translation deltas propagated to descendant RICs — and since the fast path clears the dirty flags, the assumed "re-sync on settle" never happened. The post-release forced re-raster then drew the layer's entity range from stale descendant RICs, permanently baking the pre-gesture child positions while the DOM, source, and overlay all carried the resize. Fixed by propagating the world-space delta to descendant RICs for all delta shapes (the math is affine-general). Reproduced red→green at two levels: a compositor golden probing the settle raster directly, and a full-editor gl_rnr replay asserting presented content tracks a corner resize mid-drag and after settle.

## Testing

`TextTool_tests.cc` (30 cases: gestures, caret placement incl. transformed text, undo labels, soft-wrap round-trip, ConcurrentDom scoped-access regression), `SelectionAabb`/`OverlayRenderer`/`SelectTool` text-chrome tests, `CanvasScrollbars_tests.cc`, compositor goldens for group transforms + the settle re-raster, and gl_rnr replays driving the full shell event loop (double-click typing via live ImGui pointer and document-space input, high-zoom coverage, group first-resize lockstep). Full `bazel test //...`: 722 pass, 0 failures.

https://claude.ai/code/session_01SgBwwd47scQ8vXhKXXq8rc


jwmcglynn:
Approved

Issue #149, opened 2024-05-20
Project Status


jwmcglynn:
# Readiness

Latest release [v0.5.0](https://github.com/jwmcglynn/donner/releases/tag/v0.5.0)

- Implements most static SVG functionality, including `<filter>`, `<text>`, and more!

Next on roadmap: [v1.0: Interactive editing, conformance, new GPU-accelerated rendering backend](https://github.com/jwmcglynn/donner/milestones/3).

# W3C Spec Versions

https://www.w3.org/TR/2018/CR-SVG2-20181004/

# SVG Features

- [x] Basic shapes
- [x] Gradients
- [x] Fills
- [x] Strokes (and dash patterns)
- [x] CSS3 Stylesheets
- [x] Patterns (#4)
- [x] Text (#242)
- [x] Image support (#187) 
- [x] Marker support (#185)
- [x] Masking (https://www.w3.org/TR/css-masking-1/) #186 
- [ ] Compositing (https://www.w3.org/TR/compositing-1/)
- [x] Filter effects (https://www.w3.org/TR/filter-effects/)
  - [x] #151  
  - [x] feBlend
  - [x] feColorMatrix
  - [x] feComponentTransfer
  - [x] feComposite
  - [x] feConvolveMatrix
  - [x] feDiffuseLighting
  - [x] feDisplacementMap
  - [x] feDropShadow
  - [x] feFlood
  - [x] feGaussianBlur
  - [x] feImage
  - [x] feMerge
  - [x] feMorphology
  - [x] feOffset
  - [x] feSpecularLighting
  - [x] feTile
  - [x] feTurbulence
  - [x] feDistantLight
  - [x] fePointLight
  - [x] feSpotLight
- [ ] Animation support (https://svgwg.org/specs/animations/)
- [ ] Scripting (javascript) support

# SVG Element Support

Basic shapes
- [x] `<circle>`
- [x] `<ellipse>`
- [x] `<line>`
- [x] `<path>`
- [x] `<polygon>`
- [x] `<polyline>`
- [x] `<rect>`

Rendering
- [x] `<marker>` (#185)
- [ ] `<view>`
- [x] `<filter>` (#151)

Structural
- [x] `<defs>`
- [x] `<g>`
- [x] `<svg>`
- [x] `<use>`
- [x] `<symbol>` (#281)
- [ ] `<switch>`
- [ ] `<desc>`
- [ ] `<title>`
- [ ] `<metadata>`

Styling
[...]

PR #718, opened 2026-07-03
Implement <switch> and conditional-processing attributes

jwmcglynn:
Implements the `<switch>` element and conditional-processing attributes (`requiredFeatures`, `requiredExtensions`, `systemLanguage`), closing feature gap F4 in the resvg feature-gaps catalog. Part of #149.

## Semantics (matched against resvg-test-suite goldens)

- `<switch>` renders only the first direct child whose conditional attributes all evaluate to true. Unknown (non-SVG) children and non-element nodes are never selected. `display` does not participate in selection, so a selected `display:none` child renders nothing (per `structure/switch/display-none-on-child.svg`).
- Conditional attributes on rendered elements outside `<switch>` — including the root `<svg>` and containers — disable rendering of the element and its subtree when false. Non-rendered referenced elements (gradients, `<clipPath>`, `<defs>` content) are unaffected, matching resvg's `on-clipPath` / `on-defs` / `on-linearGradient` goldens.
- `<tspan>` honors conditional attributes through TextSystem span hiding.
- Evaluation rules: `requiredFeatures` always evaluates true (deprecated in SVG2; verified against `structure/switch/requiredFeatures.svg`, which renders the child declaring a feature string); non-empty `requiredExtensions` evaluates false; `systemLanguage` matches the user language `en` by exact tag or subtag prefix (`en-GB` and `en-US` both match).

## Implementation

- New `SVGSwitchElement` DOM type (`ElementType::Switch`), registered in `AllSVGElements`; behaves like `<g>` aside from child selection (`structure/switch/with-attributes.svg`).
- New `ConditionalProcessingComponent` populated by the common attribute path in `AttributeParser`, so the attributes apply to any element; evaluation in `EvaluateConditionalProcessing` / `SystemLanguageMatches`.
- Child selection and subtree pruning in `RenderingContext` (`selectSwitchChild` + a conditional check during render-tree traversal, mirroring `display:none` pruning); tspan handling in `TextSystem`.

## Tests

Red→green on this branch (red run captured in the first commit):

- `StructureSwitch/*` — 12 previously-skipped tests un-skipped, all were red at 160,000 px differing, now green.
- `StructureSystemLanguage/*` — `on-svg` (161,980 px), `on-tspan` (9,378 px), `ru-Ru` (160,000 px) red→green; the 7 already-passing tests in the category stay green.
- `masking/clipPath/switch-is-not-a-valid-child.svg` — un-skipped; passes before and after the implementation (a `<switch>` is not a valid clipPath child and contributes no clip geometry).

All of the above verified on default (tiny-skia / simple text), `--config=text-full`, and `--config=geode` (both `TinyGolden` and `GeodeGolden` modes), with default pixel budgets — no thresholds added.

New unit tests:

- `donner/svg/components/tests/ConditionalProcessingComponent_tests.cc` — systemLanguage matching (exact, subtag, list, case-insensitivity, empty) and attribute evaluation rules, via named gmock matchers.
- `donner/svg/tests/SVGSwitchElement_tests.cc` — DOM casting plus ASCII-render coverage for child selection, requiredExtensions skipping, requiredFeatures pass-through, systemLanguage selection, unknown-child skipping, display:none selection semantics, group-attribute application, and conditional pruning outside `<switch>`.

The F4 section is removed from `docs/design_docs/0021-resvg_feature_gaps.md` and the derived skip count refreshed. `python3 tools/cmake/gen_cmakelists.py --check --build` passes.

https://claude.ai/code/session_01Rk85ZkdhJ4a2pVRtGbjpgt


jwmcglynn:
Approved

PR #782, opened 2026-07-06
Editor polish wave 1: text tool UX, transform editor, perf overlay, layers/canvas fixes

jwmcglynn:
Integration umbrella for the wave-1 editor polish packets (merged onto `qa/polish-wave1`).

- Text tool: in-canvas editing sessions with caret, box text, style toggles; click edits existing text, double-click starts point text; text picking via rendered ink; Convert Text to Outlines.
- Text frames: authored box drives the rect; rotate/resize cursor feedback and handles; rotated frames stay oriented and resize in local space.
- Transform inspector: editable decomposed numeric fields with shared field width and decomposed-edit round-trip tests.
- Perf overlay off by default with an FPS pill and grouped full-graph view; render-request counting in frame diagnostics.
- Layers panel (W10): selection/hover polish and tree-aware Bring to Front / Send to Back.
- Canvas/renderer fixes: pane owns its own scroll events, scrollbars represent the document, zoom/scroll pop fixes on viewBox-less docs, `<use>` shadow-tree and text ink-bounds render spans.

gate: editor suite green except the known /tmp sandbox target.

Part of Design 0013 / Design 0017 (zettelkasten).

https://claude.ai/code/session_01Ai8tV2bCYbDJmWk3s3Sf17

PR #795, opened 2026-07-06
Structured editing: source formatting fidelity, transform syntax preservation, undo hardening

jwmcglynn:
Grouped structured-editing packets (P1, P2, P5) merged onto `qa/polish-wave1`.

- P1: context-aware source-formatting indentation for structural insert/move in `XMLDocument`; append as a pure insertion after the last sibling, with source-formatting fidelity tests.
- P2: preserve authored transform syntax on forward writeback, with a syntax-preservation matrix and undo-bytes tests.
- P5: undo hardening: re-enable undo-across-reparse regression, skip undo entries whose rehydration target no longer resolves, preserve author display value across layers-panel hide/show.

gate: editor suite green except the known /tmp sandbox target; group branches merged cleanly with no conflicts.

Part of Design 0013 / Design 0017 (zettelkasten).

Depends on / bases on `qa/polish-wave1` (PR #782); merges the constituent packet branches, diff shown is these packets over wave-1.

### DIFF
--- a/donner/base/xml/XMLDocument.cc
+++ b/donner/base/xml/XMLDocument.cc
@@ -757,6 +757,77 @@ bool SourceHasClosingTagAt(std::string_view source, std::size_t offset,
          source.substr(offset, closingTag.size()) == closingTag;
 }
 
+// How a structural mutation should format the whitespace around a newly serialized node.
+enum class InsertionFormatting {
+  // Text content: keep it inline (e.g. `<text>hello</text>`). Never introduces newlines,
+  // matching how a human authors a text run.
+  kInline,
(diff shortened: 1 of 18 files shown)


jwmcglynn:
Triage of the Codex review (staging agent):

- Local display attribute early-return bypasses computed visibility in the layers-panel eye (P2): plausible; needs a test with display="inline" visibility="hidden".
- hiddenDisplayElements retaining SVGElement handles across deletion/reparse (P2): valid lifetime concern given ElementAnchor detached-node retention; cache should be cleared on document replacement or use non-owning targets.

Both routed to the wave-2 integration pass; no untested fixes pushed from staging.

PR #778, opened 2026-07-06
SMIL animation slice 1: animate, animateTransform, set with time-sampled rendering

jwmcglynn:
Ports the proven animation implementation from the tiny-skia branch (51068c3f, Phases 1-9, previously fuzzed to 117M+ executions) onto the current architecture, scoped to slice 1 of the declarative non-scripting target.

**Scope**: `<animate>` (scalar attributes), `<animateTransform>` (translate/scale/rotate/skew), `<set>`; from/to/by/values, keyTimes, calcMode discrete|linear; begin (offset), dur, end, repeatCount/repeatDur, fill freeze|remove, restart; sandwich composition; `SVGDocument::setTime()/currentTime()` public API. Deferred (slice 2+): animateMotion/mpath, paced/spline, syncbase, additive/accumulate, CSS animations, compositor layer caching (HintSource::Animation left dormant).

**Integration**: `AnimationSystem::advance()` runs post-`computeAllStyles`; overrides patch computed style/geometry at `Specificity::Override()`. Two architecture corrections found during the port: per-entity base-component backup so `fill=remove` reverts on live documents, and layout-cache invalidation for animated transforms.

**Tests**: 65 ported unit tests; new time-sampled golden renders (before/mid/after, freeze vs remove, repeat) across tiny-skia/text-full/geode lanes; full svg + editor suites green at t=0; 3 fuzzers ported with corpora (60s smoke: 39.5M execs on clock parsing, zero crashes).

Design doc ported with a dated slice-1 addendum recording the attach-point and scope decisions.

https://claude.ai/code/session_01WcXFtPoHgJMjdeC8rd7ZQR

PR #796, opened 2026-07-06
Fix rules_python warning and gold-linker archive order

jwmcglynn:
Fixes two small but separate issues that surfaced during the CI sprint:

- Opt into the future rules_python default with `common --incompatible_default_to_explicit_init_py`, after validating that the repo's Python targets and generated lint tests do not need implicit `__init__.py` files.
- Fix the gold-linker canary regression exposed by the SMIL merge. `_donner_perf_sensitive_cc_library_impl` was appending a library's own linking context after its deps, which can put a consumer archive after provider archives on the link line. lld tolerates that backward reference, but single-pass GNU linkers such as gold do not. Inserting the library's linking context before its deps matches native `cc_library` link order.

Verification:

- `bazel build //tools:render_notice`: rules_python warning gone.
- `bazel build -- //...`: green on branch validation.
- `bazel build -c opt //donner/editor`: green.
- `bazel test -- //tools/... //build_defs/... //tools/mcp-servers/...`: 12/12 pass.
- `bazel test --test_tag_filters=lint,banned_patterns -- //...`: 395/395 pass.
- Gold-linker root cause validated by the link-order flip and a local canary build.
- GitHub CI: hosted `linux` has passed the linker canary on `f547c500d`.

Out of scope: the analysis-cache churn between operator `-c opt` runs and RE builds remains separate.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -27,6 +27,16 @@ build --enable_platform_specific_config
 ## Build config
 ##
 
+# rules_python has deprecated implicitly creating empty __init__.py files in
+# the runfiles tree for py_binary/py_library/py_test targets (on by default
+# today, will flip in a future release). Opt in to the future default
+# repo-wide now: every py_binary/py_library/py_test in this repo (the
+# generated per-target _lint checks plus tools/* and build_defs/*) builds and
+# tests clean under this flag with no explicit __init__.py additions needed,
+# since none of these targets do package-relative imports across directory
+# boundaries. See https://github.com/bazel-contrib/rules_python/issues/2945.
+common --incompatible_default_to_explicit_init_py
+
 # Force clang to be the default compiler for all Bazel C/C++ actions.
 # Propagates CC/CXX so every action (including host tools) uses clang/clang++.
 build --action_env=CC=clang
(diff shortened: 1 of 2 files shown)


jwmcglynn:
CI note: the linux job failure here is pre-existing on main, not introduced by this PR. The Linker canary (non-lld linker support) step fails with undefined reference to donner::svg::components::AnimationSystem::advance while linking //donner/svg/tool:donner-svg, and main's own CI run for the base commit 12bf69dc (SMIL animation slice 1) fails identically: https://github.com/jwmcglynn/donner/actions/runs/28793694393. This PR changes only .bazelrc (rules_python init flag) and does not affect C++ linking. All other checks on this PR pass. If a fix for the gold-linker archive ordering lands on main, updating this branch from main should turn the job green.

jwmcglynn:
Monitor bot: the red `linux` check was the "Linker canary (non-lld linker support)" step, not the `--incompatible_default_to_explicit_init_py` change. Root cause: `_donner_perf_sensitive_cc_library_impl` appended its own linking context after its deps, so consumer archives landed after their providers on the link line; lld tolerates backward references, gold does not. #778 made `rendering_context` the sole referencer of `AnimationSystem::advance`, exposing it. Fixed in f547c500d (verified via `bazel aquery` order flip + local build of both canary targets). Note: the same canary is red on `main` and on #799; this commit needs to reach `main` to heal those.

PR #808, opened 2026-07-10
Binary size: baseline report + wasm size tooling

jwmcglynn:
## What

Makes binary size a measured, first-class property and records a reproducible baseline. Tooling and docs only; no functional or API changes.

- Extend `tools/binary_size.sh` to also build and measure the tiny_skia WebAssembly renderer (`//donner/svg/renderer/wasm:donner_wasm`) alongside the native binaries. It now reports raw and gzip-9 transfer sizes for the `.wasm` and JS glue, plus a bloaty section/symbol breakdown of the `.wasm`. `SKIP_WASM=1` (or a host without emcc) skips the wasm section.
- New design doc `docs/design_docs/0055-binary_size.md` records the baseline and a ranked reduction plan.

## Measured baseline (an internal build host, macOS arm64)

Native, `--config=macos-binary-size`, stripped:

| Target | Size |
| --- | --- |
| svg_parser_tool.stripped | 2,527,384 bytes (2.41 MiB) |
| donner-svg.stripped | 4,182,688 bytes (3.99 MiB) |

WebAssembly, `--config=wasm` (current default; note it builds unoptimized at fastbuild -O0 with no wasm-opt):

| Artifact | Raw | gzip -9 |
| --- | --- | --- |
| donner_wasm_bin.wasm | 5,688,030 | 1,184,620 |
| donner_wasm_bin.js | 154,601 | 41,847 |
| Total transfer | 5,842,631 | 1,226,467 |

## Regenerate

```
bash tools/binary_size.sh
```

## Follow-on

Reduction PRs land on top of this. The largest measured lever is building the shipped wasm with -Oz (see doc 0055 for the full ranked plan and per-variant measurements).

### DIFF
--- a/docs/design_docs/0055-binary_size.md
+++ b/docs/design_docs/0055-binary_size.md
@@ -1,7 +1,7 @@
 # Design: Binary Size Reduction
 
 **Status:** Draft
-**Author:** Claude Opus 4.8
+**Author:** Claude Fable 5 (reviewed; drafted by Claude Opus 4.8)
 **Created:** 2026-07-10
 
 ## Summary
@@ -87,7 +87,7 @@ file. The wasm target builds under `--config=wasm` and is measured both raw and
 gzip -9 (the transfer-size proxy), with a bloaty section/symbol breakdown of the
 `.wasm`.
 
-## Baseline (measured 2026-07-10, deep-thought, macOS arm64)
+## Baseline (measured 2026-07-10, an internal build host, macOS arm64)
 
 Native, `--config=macos-binary-size`, stripped, arm64 Mach-O:
 
--- a/docs/design_docs/0055-binary_size.md
+++ b/docs/design_docs/0055-binary_size.md
@@ -1,0 +1,167 @@
+# Design: Binary Size Reduction
+
+**Status:** Draft
+**Author:** Claude Opus 4.8
+**Created:** 2026-07-10
+
+## Summary
+
+Donner should build fast, stay small, embed easily into apps, and download
+cheaply as WebAssembly. This design makes binary size a first-class, measured
+property: it records a reproducible baseline for the representative
+native and wasm artifacts, extends the checked-in tooling so anyone can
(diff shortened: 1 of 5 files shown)


jwmcglynn:
CI fix pushed (22ff4752): raised timeout-minutes from 60 to 120 for exactly the linux-self-hosted and coverage-self-hosted jobs.

Evidence: both lanes were cancelled at exactly the 60-minute job limit on both attempts of the previous head (1h0m21s each, original run + one rerun), every time mid Build / Generate coverage with no build or test failure reported. The same 60-minute cancellation pattern hit unrelated branches in the same window, which points at slot contention from concurrent coverage-class builds on the shared self-hosted queue, not at this change.

Scope: only the two timeout-minutes values (plus their adjacent comments) changed. Gating, skip logic, and the RE fast-fail (no local fallback) behavior are untouched. Watching the fresh run to confirm the new jobs actually carry the 120-minute limit.

jwmcglynn:
Rebased onto current main. The earlier commit that raised the linux-self-hosted and coverage-self-hosted job timeout to 120 minutes has been dropped: that timeout raise was a stopgap for queue contention, and the merged per-lane serialization addresses the root cause, with lanes now completing well under the original 60-minute limit. This branch therefore no longer touches .github/workflows at all and is limited to tools/ and docs/design_docs/0055-binary_size.md. Local gates pass: bash -n on tools/binary_size.sh and //tools:binary_size_tests.

jwmcglynn:
CI status after rebase: all standard lanes green. linux-self-hosted passes in 2m48s once it runs uncontended (earlier reds on the self-hosted lanes were transient concurrency preemption while other runs held the shared self-hosted RE slots, not failures in this change, and 2m48s is well under the original 60-minute limit). coverage-self-hosted fails deterministically at the Generate coverage step with no executable line data because this branch touches only tools/ and docs/ and adds no instrumentable C++, so the coverage report is empty; this is independent of the rebase and the PR is MERGEABLE (the coverage lane is not a required check).

PR #809, opened 2026-07-10
Geode: GPU transfer-volume counters + measured cost profile in design 0030

jwmcglynn:
## Summary

First slice of a Geode performance pass: make GPU transfer volume measurable, and record a measured steady-state cost profile in design doc 0030.

- Adds three counters to GeodeCounters: bufferWrites, bufferWriteBytes (payload bytes of every wgpu::Queue::writeBuffer), and textureWriteBytes (writeTexture payloads), wired via GeodeDevice count helpers at every upload site in GeoEncoder, GeodeTextureEncoder, and GeodeFilterEngine. Null-counter cost unchanged (one pointer compare per site).
- GeodePerf_tests logs + RecordPropertys the new counters per fixture.
- 0030 gains an appendix, Measured steady-state cost profile, with the numbers below and a cost ranking; stale pre-0041 text (MSAA stays 4x boundary, alpha-coverage fallback non-goal, impl_->msaaTarget, deleted alpha-coverage shader variant) is removed. Also fixes the stale MSAA comment on GeodeDevice::sampleCount().

## Measurements (macOS/Metal, Apple M4 Pro, headless; bazel test //donner/svg/renderer/geode:geode_perf_tests --test_output=all --nocache_test_results)

Steady state = second render of an unchanged document (frame 2):

| Fixture | draws | bufferWrites | bufferWriteBytes | bindgroupCreates | bufferCreates |
|---|---|---|---|---|---|
| SimpleShapes | 3 | 24 | 4,128 | 3 | 9 |
| Moderate | 3 | 17 | 5,372 | 3 | 18 |
| lion.svg | 132 | 1,056 | 172,776 | 132 | 12 |
| Ghostscript_Tiger | 304 | 2,432 | 1,473,888 | 304 | 20 |

Headline: a completely static Tiger re-uploads ~1.44 MB across 2,432 writeBuffer calls every frame (8 writes per draw; ~88 MB/s at 60 fps) because cached EncodedPath data has no persistent GPU residence. Bind groups are recreated per draw per frame. These are the optimization targets for the follow-up PRs.

## Testing

- bazel test //donner/svg/renderer/geode/... //donner/svg/renderer/tests:renderer_geode_golden_tests : green (23 passed, 6 sharded impls skipped as usual).
- No behavior change: counters are passive instrumentation; goldens untouched.

### DIFF
--- a/docs/design_docs/0017-geode_renderer.md
+++ b/docs/design_docs/0017-geode_renderer.md
@@ -792,6 +792,33 @@ struct GeodeLayerStack {
 - **Patterns:** Render the pattern tile into a texture, then use it as a repeating shader in
   subsequent draws.
 
+### Debug and Diagnostics
+
+Geode ships a geometry debug overlay for inspecting the banded geometry
+it emits per draw:
+
+- `RendererGeode::setDebugGeometryOverlay(bool)` (surfaced as a
+  default-no-op virtual on `RendererInterface` and forwarded by the
+  `svg::Renderer` facade) makes every path draw additionally outline its
+  Slug band decomposition: horizontal band strips (cyan), vertical band
+  strips (yellow), and the whole-path bounding quad with its triangle
+  diagonal (magenta) - the two triangles Geode actually rasterizes per
+  path. Overlay geometry renders as hairline EvenOdd rings through the
+  existing fill pipeline; no dedicated shaders or pipelines exist for it.
+- Default off. When off, rendering is byte-identical to a build without
+  the feature (asserted by `GeodeDebugOverlay_tests` and the geode
+  goldens); the only cost is one boolean test per draw. `<use>`
+  instancing is disabled while the overlay is on so overlay draws stay
+  in paint order.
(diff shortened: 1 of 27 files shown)


jwmcglynn:
CI note: linux-self-hosted and coverage-self-hosted were cancelled at the 60-minute job limit on all three attempts (original + two reruns), each time mid Build / Generate coverage with no build or test failure reported. The same cancellation pattern hit unrelated branches around the same window (e.g. size/binary-size-pass 07:32Z, codex/wasm-editor-parity 04:54Z), which points at self-hosted runner saturation rather than this change. Note the branch-push and stacked-PR runs skip these two jobs via gatekeeper, so no run of this commit has executed them to completion yet.

Coverage that did run is green: linux (tiny_skia), macos (tiny_skia), macos-self-hosted, asan-geode (14m31s), lint, cmake-validate; locally on an internal build host the geode suite, geode goldens, renderer_tests, the resvg geode suite, and the full editor suite all pass fresh. Not re-running further to avoid adding load to the contended queue; happy to rerun once the runner frees up.

PR #906, opened 2026-07-19
Stabilize Geode editor rendering and Wasm integration

jwmcglynn:
This is the first of four dependency-ordered Geode optimization changes. It stabilizes the editor rendering path across native, headless, and WebAssembly configurations.

## Changes

- keeps resident GPU buffers and WebGPU callback state alive for the full submission lifetime
- supports explicit Geode texture snapshot readback and rendering without an externally supplied device
- fixes document-access, compositor, editor coordination, and renderer move-lifetime defects with regression coverage
- hardens WebAssembly builds and browser isolation checks across supported browser configurations
- adds Geode-specific editor goldens and serializes tests that require an exclusive hardware adapter

## Validation

- Final cumulative stack gate: `tools/llm-bazel-wrap.sh test --config=geode //... --test_env=DONNER_GEODE_FORCE_FALLBACK_ADAPTER=1 --test_output=errors --noshow_progress --show_result=0`
- Cumulative result: 868 passed, 8 declared platform skips, exit 0
- Focused renderer move-assignment and lifetime tests, lint checks, browser smoke coverage, and exact editor goldens pass
- The complete gate uses the deterministic software Vulkan adapter. Hardware-adapter stress remains a separate environment-specific validation boundary.

## Security review

- no new network service, external interface, persistence mechanism, or privilege boundary
- dependency revisions remain pinned, and local patches are limited to compiler and platform portability
- renderer inputs remain covered by unit, golden, fuzz, browser, and repository-wide suites
- secret and publication-scope scans are clean

Stack: 1 of 4. The next PR adds ray pruning, sorted traversal, and adaptive band selection.

PR #923, opened 2026-08-07
Bring the Wasm editor to release quality: presentation fixes, shared seams, Geode-only cleanup

jwmcglynn:
This PR brings the browser (WebAssembly) editor to release quality for v0.8. It replaces the split main-thread/worker presentation with a single-canvas whole-app-in-worker architecture, fixes every user-visible defect found in three-browser validation (pinch parity, zoom flicker, blank pane on load, wrong z-order for selection chrome, missing cursors, a boot deadlock, and a curve-fidelity violation), deletes the dead dual-backend tier and roughly 8,000 lines of dual-path presentation code, and overhauls the browser test harness so its pixel probes measure real composited output. The branch carries the editor stabilization line rebased onto main and reconciled with the Geode stabilization work in #906.

## Architecture: the whole app runs in a worker

The editor previously ran ImGui on the main thread while a render worker presented the document through a second, worker-owned canvas. That split was the root of a class of presentation bugs (z-order, atomicity, gesture flicker) and doubled the presentation code. Now the entire application runs on a worker pthread against one transferred OffscreenCanvas:

- **Demand-driven frame loop.** The app thread ticks on animation frames but runs a frame only when something asked for one (editor request, browser input, or a due idle timer). An idle editor parks the loop and runs zero frames; input wakes it on the next tick.
- **Page/worker bridge.** DOM state the app needs every frame (canvas size, device pixel ratio, input modifiers) crosses through a small shared-memory mirror; DOM writes (cursor changes, stats publishes) proxy to the main thread asynchronously. Diagnostics used by the test suites publish through the same bridge.
- **One presentation path.** Document tiles cross the thread boundary as CPU bitmaps and are uploaded into texture snapshots owned by the presentation cache, so the same Geode framebuffer underlay that desktop uses presents them beneath ImGui chrome. Selection chrome, bounds rects, and overlays now always render above the document. The ImGui image blit survives only for non-WebGPU builds and content-only capture.
- **Correct alpha end to end.** Uploaded CPU bitmaps are straight-alpha and are tagged as such on their snapshots; the texture draw path converts in the shader instead of claiming premultiplied storage, which would have fringed every antialiased edge.

## User-visible fixes

- **Selection chrome z-order.** Path overlays and bounds rects rendered beneath the document on the browser build; the underlay unification above fixes the layering structurally rather than patching draw order.
- **Trackpad pinch zoom parity.** Desktop applies `zoom = 1 + magnification` exactly; the Safari gesture bridge used an ad-hoc constant that netted about 2.1x too slow in log space. Pinch conversion now lives in one policy header, the derived bridge constant is published to the page at startup, and browser tests assert the applied zoom magnitude within 2 percent. Browsers that deliver pinch as ctrl+wheel also get preventDefault applied from a non-passive listener mirrored through the bridge, so the page no longer browser-zooms during canvas pinches.
- **Zoom flicker and blank pane on load.** Presentation geometry and pixels change atomically against the viewport each frame was rasterized with, and the presenter holds the last accepted placement across document swaps, bounded by a frame budget.
- **Cursors.** Cursor changes are DOM writes; they now proxy to the main thread and target the real canvas element, restoring resize/rotate/tool cursors.
- **Boot deadlock.** The desktop frame loop blocks in waitEvents; inside a browser animation-frame tick that blocked the only thread that could deliver the wake. The wait is now desktop-only and the browser loop parks by declining ticks.
- **Curve fidelity.** Stroke flattening tolerance was a fixed path-local constant, so deep zooms rendered circles as visible line segments. Tolerance is now derived from the device transform (about a quarter device pixel, bucketed by power-of-two scale, refine-only) and participates in the stroke cache key; `Path::strokeToFill` no longer has a default tolerance, making the choice a compile-time obligation. An enforcement suite pins the invariant.
- **Faster boot.** The browser build was acquiring a headless WebGPU device on the UI thread that nothing ever rendered with; removing it cuts first frame from 3.2 s to 1.9 s on the software-rasterizer lane.

## Test harness and coverage

- **Pixel probes now see the canvas.** Headless Chromium's default GL path never composited the worker's OffscreenCanvas swapchain into `page.screenshot()`, so every pixel probe in the suite read the flat page background. Routing GL through ANGLE (non-Linux) makes the probes real; this alone resolved five failures that were measuring nothing and cut suite wall time from 1.7 m to 41 s. The Linux lane pins its own GPU arguments and needs the equivalent validated there (tracked follow-up).
- **Frame-loop contracts retargeted.** The old presentation-only-frames and idle-callbacks tests asserted the two-canvas architecture and had become unsatisfiable by construction. Their replacements assert the demand-driven loop's invariants (bounded input-attributed wakes, park-to-zero-frames, wake-on-demand with declined-tick liveness) and were each verified to fail under two distinct sabotages before landing.
- **Perceptual golden comparison.** Byte-identical golden comparison for browser captures is replaced with a pixelmatch-based perceptual diff, eliminating cross-host antialiasing churn while still catching real regressions.
- **Readback diagnostics.** An opt-in URL parameter publishes WebGPU readback statistics (attempts, retries, rescues) through the bridge for debugging capture issues in the wild.

## Verification

- Desktop: editor, renderer, and compositor suites green (258 editor test targets; full sweep).
- Browser (headless Chromium, ANGLE): smoke, presentation-regression, and both composited-invariant lanes green except four tracked failures, each root-caused with evidence and waiting on a maintainer decision rather than a diagnosis: two assert the pre-prewarm selection contract, one has a stale probe-region inset, and one is a background-predicate collision (details in the tracked follow-ups).
- Manual three-browser validation: Chrome and Safari at 60 fps with all tested functionality working; Firefox good with occasional long render frames during shape drags (tracked follow-up).
- Wasm gates: package size, function-body budget, and transition tests pass.

## Follow-ups (tracked, not in this PR)

- Selection-prewarm test contract: assert the intentional single prewarm render or reinstate the browser disable.
- Trackpad-pinch probe geometry recalibration (stale pane inset clips the measured glyph).
- Firefox long render frames during shape drags.
- Linux browser lane compositor validation (ANGLE equivalent), issue #927 sampling differences.
- WebKit bitmap-bridge keep-or-retire decision.
- Worker-surface transparency checkerboard successor (destination-over pass) and the remaining items in #924, #925, #926.


donner/editor/wasm/tests/wgpu-readback-contract.spec.mjs
```diff
@@ -0,0 +1,356 @@
...
+});
+
+test("direct worker surfaces cross a local event-loop task before acceptance", () => {
+  const taskCompletion = workerRendererSource.match(
+    /void AsyncRenderer::runWorkerTask\(void\* [^)]*\)([\s\S]*?)\n}\n\nvoid AsyncRenderer::acknowledgeDirectSurfaceTaskBoundary/,
+  );
+  assert.ok(taskCompletion, "expected the worker task-boundary handoff");
+  assert.match(
+    taskCompletion[1],
+    /emscripten_set_timeout\(&AsyncRenderer::acknowledgeDirectSurfaceTaskBoundary,/,
```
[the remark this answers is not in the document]
jwmcglynn: Fixed in a1e7fc274 (this review examined an earlier head): the contract now pins the MessagePort scheduler — it requires ScheduleWorkerTaskBoundaryCallback with the MessageChannel macrotask hop and explicitly rejects timer hops, which inherit the caller's nesting level and hit the 4 ms nested-timer clamp.

build_defs/rules.bzl
```diff
@@ -188,17 +188,82 @@ def _donner_transitioned_executable_impl(ctx):
...
+    *":$candidate:"*) ;;
+    *) library_path="${{library_path:+$library_path:}}$candidate" ;;
+  esac
+}}
+
+if [[ -n "$runfiles_dir" && -e "$runfiles_dir/$logical_impl" ]]; then
+  impl="$runfiles_dir/$logical_impl"
+  for candidate in "$runfiles_dir"/*/_solib_*/*; do
+    if [[ -d "$candidate" ]]; then
+      append_library_dir "$candidate"
```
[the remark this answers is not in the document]
jwmcglynn: Good catch — fixed: the directory-runfiles branch now appends the parent directory of flat-layout .so/.dylib entries, mirroring the manifest branch, so LD_LIBRARY_PATH/DYLD_LIBRARY_PATH is populated for both _solib layouts.

docs/design_docs/0056-wasm_ios_runtime_compatibility.md
```diff
@@ -1,130 +1,91 @@
-# Design: WASM iOS Runtime Compatibility
```
jwmcglynn, line 1: Looks like we clobbered the 0056 design and did not rename it? Let's rename this to match the current title and update the index

donner/svg/renderer/RendererInterface.h
```diff
@@ -472,10 +565,54 @@ class RendererInterface {
...
+   * compose instead. Fully-opaque pixels are unchanged, fully-transparent
+   * pixels become checkerboard, and partial alpha blends as `destination-over`
+   * does. Call it after the last `endFrame()` of the frame and before handing
+   * the target to the presenter.
+   *
+   * @param params Checkerboard placement and appearance.
+   * @return True when the pass was submitted. Backends without a GPU frame
+   *   target return false and leave the target untouched.
+   */
+  virtual bool drawCheckerboardUnderlay(const CheckerboardUnderlayParams& /*params*/) {
```
jwmcglynn, line 612: I'd like to find another way to do this that doesn't pollute RendererInterface - is there a helper we can make that can work with both geode and SW without changing RendererInterface?

donner/svg/renderer/Renderer.h
```diff
@@ -110,6 +162,17 @@ class Renderer : public RendererInterface {
...
 
+  /**
+   * Renders an element subtree into a directly sampleable backend texture.
+   *
+   * This is the GPU-native counterpart to \ref renderElementToBitmap. It uses
+   * the same bounds, crop, and isolated offscreen draw, then transfers
+   * ownership of the rendered target instead of reading it back through the
+   * CPU. Backends without texture snapshots return nullptr.
+   */
+  [[nodiscard]] std::shared_ptr<const RendererTextureSnapshot> renderElementToTextureSnapshot(
```
jwmcglynn, line 173: Can we combine this with relevantElementToBitmap somehow? Return a wrapper object that can either have the bitmap or texture handle, and have an API that does the readback to get the pixels if necessary?

Does the user need to know its in a texture or can we just shuttle around a wrapper that handles both behaviors?

jwmcglynn: Combined in #929: `renderElementToBitmap` and `renderElementToTextureSnapshot` are now one `Renderer::renderElement` returning `svg::RendererImage`, a wrapper holding either the backend texture or CPU pixels, with `bitmap()` doing the readback once on demand. Callers no longer ask which backend they are on: LayersPanel stores one `RendererImage` per row and the ImGui upload path picks the texture provider or the bitmap provider off `textureSnapshot()`. The duplicated document-prepare/crop body between the two old entry points is gone, and `RendererInterface::supportsElementTextureSnapshots()` went with it, since the choice is now just whether `textureSnapshot()` is non-null.

donner/svg/renderer/RendererInterface.h
jwmcglynn: Done in #929, though not as a both-backends helper, because there is no software consumer to write one for: the checkerboard exists so see-through document pixels reach a GPU presentation surface with something behind them, and Geode was the only implementation. `drawCheckerboardUnderlay` is deleted from RendererInterface, Renderer, and RendererGeode; both copies of the draw body collapse into one `geode::GeodeCheckerboardPass` next to the pipeline they shared, owning the per-consumer uniform buffer and bind group (nothing allocates per frame) and taking a scissor so the editor still clips to the render pane. The params and appearance constants move into `donner::geode` with it, and the destination-over coverage moved onto the pass with a new scissor case.

PR #931, opened 2026-08-13
Cut CI test-suite runtime: unblock the queued GPU stall, halve the suite's CPU, and make stalls diagnosable

jwmcglynn:
The self-hosted CI lanes were taking hours and dying at their 210-minute backstop. Almost none of it was compute. Mining the CI diagnostics artifacts from four runs: the whole test phase resolves in about 60 seconds, then Bazel parks on a single queued action for ~200 minutes. Three geode suites requested a remote-execution worker property no worker in the pool advertises, and a platform property is a hard match with nothing to bound the wait - a per-test timeout only starts once a test *runs*. Every test queued behind that point never ran. (That fix is already on main via #929.)

Underneath it, the suite's intrinsic cost was mostly scaffolding. Of 2297 test CPU-seconds, 714 (31%) were 476 per-target lint py_tests running a regex scan that takes 2.5 seconds for the whole repository in one process, and 576 (25%) were fuzz timers mutating for 10 seconds each in a build with no sanitizers. The product tiers were fine: the base tier is 3.4 minutes.

This branch:

- Scopes `exclusive-if-local` to the tests that actually open a GPU device. It was applied to every geode-*backend* target including the `text_full` tier, putting 92 targets into Bazel's single-file exclusive tail; now 31.
- Stops discarding the analysis cache between CI's build and test steps, and gives the coverage lane the same `--config=ci` its own fetch step uses - without it, that lane compiled a different configuration and shared no action-cache entries with the CI test lane.
- Replaces the 476 lint tests with one repo-wide `tools/lint.sh` over a strictly larger file set.
- Splits the fuzz budget so exploratory mutation happens in the nightly sanitized lane, longer than before, instead of on every unrelated PR.
- Cuts loop counts that cannot observe anything the first iterations do not, and normalizes a 10 ms poll quantum to 1 ms at identical deadlines.
- Fixes coverage's "cover everything" resolving to "cover nothing" on `build_defs/*` changes, and makes that class of drift impossible by publishing one routing decision.
- Turns the GPU inventory from a file-hash tripwire into a semantic ratchet that runs under `bazel test //...`.
- Makes a stuck target diagnosable: truncated-profile tolerance in the diagnostics report, a "configured but never finished" section, and a watchdog that names the target and fails after 15 minutes of no progress instead of 210 minutes of silence.

Measured on an 8-core machine with every test forced to execute: the full suite is 405 targets / 1250 CPU-seconds / 304 s wall, down from 889 targets.


donner/base/fonts/BUILD.bazel
```diff
@@ -94,3 +95,24 @@ donner_cc_fuzzer(
         ":woff2_parser",
     ],
 )
+
+# Source inputs for the GPU inventory ratchet (//tools/gpu_inventory). The
+# manifests are derived from a lexical scan of these files, so they must be
+# declared as real Bazel inputs for the check to be a cacheable test with
+# correct change detection. See tools/gpu_inventory/README.md.
+filegroup(
```
jwmcglynn: Let's DRY this and put this into a .bzl file as a way to export GPU srcs

donner/editor/tests/AsyncRenderer_tests.cc
```diff
@@ -1208,12 +1214,12 @@ TEST(AsyncRendererTest, DisplayNoneSelectionDoesNotLeaveStaleBackgroundPixelsWhe
...
     // WaitForRenderResult).
-    for (int i = 0; i < 3000; ++i) {
+    for (int i = 0; i < 30000; ++i) {
       auto result = asyncRenderer.pollResult();
       if (result.has_value()) {
         return result;
       }
-      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
```
jwmcglynn: Can we DRY this sleepwait too?

donner/svg/components/filter/BUILD.bazel
```diff
@@ -107,3 +107,24 @@ donner_cc_test(
         "@com_google_gtest//:gtest_main",
     ],
 )
+
+# Source inputs for the GPU inventory ratchet (//tools/gpu_inventory). The
+# manifests are derived from a lexical scan of these files, so they must be
+# declared as real Bazel inputs for the check to be a cacheable test with
+# correct change detection. See tools/gpu_inventory/README.md.
+filegroup(
```
jwmcglynn: After we DRY this we should reduce the comment in each instance - this is duplicating a lot of lines.  We may even want to roll this into a per-BUILD file helper so we can shove other global rules into it as well (like lint and doxygen verifiers)

tools/gpu_inventory/BUILD.bazel
```diff
@@ -53,3 +58,86 @@ py_test(
...
+    size = "small",
+    srcs = [
+        "generate_gpu_manifests.py",
+        "manifest_freshness_tests.py",
+    ],
+    data = [
+        "rust_allowlist.json",
+        ":manifests",
+        "//:gpu_inventory_srcs",
+        "//build_defs:gpu_inventory_srcs",
```
jwmcglynn, line 82: Is there any way we can glob these automatically with a genrule?

donner/base/fonts/BUILD.bazel
jwmcglynn: Done in 945350f45: build_defs/package.bzl exports donner_package(), one line per call site. Net 1192 lines deleted for 142 added across 61 BUILD files. Five non-first-party packages (root, build_defs, third_party trio) pass explicit globs so vendored stb headers and the WebGPU stub stay out of the inventory scan. Verified identical: bazel query on the filegroups gives the same 61 targets and 1353 files before and after. The identity check also caught the root glob following bazel-* symlinks into the output base (27010 matches vs 1353 real files); the macro excludes output trees so that fix lands once.

donner/svg/components/filter/BUILD.bazel
jwmcglynn: Done in 945350f45: the rationale lives once at the macro definition; call sites carry no comment, just donner_package(). The docstring names the growth path explicitly: a per-package lint target or doxygen verifier becomes a keyword argument here and every package inherits it without touching 60 BUILD files, and it states that none of that is wired yet so the seam does not read as load-bearing.

tools/gpu_inventory/BUILD.bazel
jwmcglynn: Structurally no: glob() is package-local by construction (stops at any BUILD boundary), and a genrule cannot escape it either since its inputs must be labels some package already exposes. Menu considered: (a) the one-line macro, shipping - a package that forgets it fails loudly via test_every_inventoried_file_is_a_declared_input; (b) repo-level sh_test running git ls-files at test time - rejected, reintroduces the non-hermetic git dependency and is invisible to target selection; (c) module-level generated hub - heavy for the gain; (d) a Starlark aspect collecting srcs/hdrs from root targets - the one real alternative, no declarations and cannot drift, but misses non-srcs files (patches, MODULE.bazel, .bzl/BUILD - most of what the Rust ratchet reads) and costs a custom rule+aspect vs a seven-line macro. Worth revisiting if the per-package line becomes a maintenance burden.

donner/editor/tests/AsyncRenderer_tests.cc
jwmcglynn: Done in 26123fb1c, and the shape repeated 42 times across the file, differing only in bound and predicate. Four helpers replace them (PollForResult iteration- or deadline-bounded, PollUntil, WaitUntil): 131 insertions, 260 deletions. The iteration-bounded overload is kept deliberately - re-expressing those as wall-clock deadlines would silently tighten timeouts on loaded runners. One wait stays unbounded with a comment: deadlining the post-cancelInFlight drain would let the test assert against a half-cancelled renderer, converting a hang into a confusing failure. Sibling files share the same shape; hoisting the helpers into a test-support header is a follow-up to keep this diff contained.

PR #939, opened 2026-08-14
Expand security analysis and fuzzing gates

jwmcglynn:
Expands the security CI surface across desktop and Wasm, bounds WOFF2 decoder work, and fixes rendering and editor CI regressions exposed by the new gates.

- runs security-extended CodeQL on pull requests and extracts the main desktop, text-full, Geode, and Wasm entrypoints
- exercises existing fuzzers under default, text-full, and Geode configurations when fuzz-relevant source changes
- replays existing parser and renderer corpora under UBSan and runs the shipped Wasm render boundary test
- scopes a known `_FORTIFY_SOURCE` macro-redefinition suppression to external Brotli so the text-full WOFF2 lane compiles under `-Werror`
- validates the WOFF2 directory and collection structure before decoder entry, writes final output lazily, and independently bounds compressed input, intermediate data, transformed-glyph scratch work, tables, fonts, and table references
- phases the Editor Wasm drag smoke test around app-consumed input and gives every fresh full-canvas capture a unique texture generation so high-zoom renders cannot alias stale pixels
- keeps the macOS GLFW runtime alive across repeated editor windows, waits on thumbnail completion with a bounded wall-clock deadline, and trims unused interaction telemetry to preserve the packaged-JavaScript size gate

Validation:
- workflow YAML parsed and repository formatting checks passed
- all referenced targets resolved in their exact Bazel configurations against current main
- reproduced the macOS text-full WOFF2 fuzzer failure without the Brotli suppression, then passed the exact fuzzer and soak command with it
- restored the former eager WOFF2 output allocation as a negative control: the exact ASan soak hit its two-second per-input watchdog, while the candidate fuzzer and soak passed uncached
- accepted a generated valid font with a 5 MiB non-`glyf` table while rejecting transformed-`glyf` and aggregate intermediate data at their exact cap-plus-one boundaries
- passed the repaired drag smoke test 20 consecutive times, the unchanged high-zoom pixel storm 10 consecutive times, and the complete five-lane browser matrix
- passed the macOS editor shell test 80 consecutive times at both 4-way and 14-way parallelism
- passed the Editor Wasm package-size target at 48,287 bytes against the 48,300-byte gate
- the exact 21-file diff passed `git diff --check`


.github/workflows/fuzz.yml
```diff
@@ -81,6 +93,30 @@ jobs:
...
             //...
 
+      - name: Test text-full fuzzers
+        run: |
+          bazelisk test \
+            --config=asan-fuzzer \
+            --config=fuzz-soak \
+            --config=text-full \
+            --test_output=errors \
+            //donner/base/fonts:woff2_parser_fuzzer \
```
jwmcglynn, line 103: I'd like to avoid explicitly naming targets in .yml files, can this use labels to filter?

.github/workflows/codeql.yml
```diff
@@ -3,6 +3,7 @@ name: "CodeQL"
 on:
```
jwmcglynn, line 3: How important is codeql? I'm concerned about another build reducing time-to-signal on PRs, maybe we remove this entirely...?

.github/workflows/codeql.yml
```diff
@@ -74,7 +93,25 @@ jobs:
       - if: matrix.build-mode == 'manual'
         shell: bash
         run: |
-          bazel build --spawn_strategy=local --nouse_action_cache --noremote_accept_cached --noremote_upload_local_results --disk_cache= //donner/svg/tool
+          bazel build --spawn_strategy=local --nouse_action_cache --noremote_accept_cached --noremote_upload_local_results --disk_cache= \
+            //donner/base/xml:xml_tool \
```
jwmcglynn, line 97: Can we use labels to select these targets instead of enumerating them all here?

.github/workflows/codeql.yml
```diff
@@ -3,6 +3,7 @@ name: "CodeQL"
 on:
   push:
     branches: ["main"]
+  pull_request:
```
jwmcglynn, line 6: At minimum I don't want this on pull request

.github/workflows/fuzz.yml
```diff
@@ -28,13 +28,25 @@ on:
...
     paths:
+      - ".bazelrc"
+      - "MODULE.bazel"
+      - "MODULE.bazel.lock"
       - "**/*_fuzzer.cc"
       - "**/*_fuzzer.cpp"
       - "**/*_fuzzer.h"
       - "**/corpus/**"
-      - "build_defs/rules.bzl"
+      - "build_defs/**"
```
jwmcglynn, line 38: Wouldn't this reduce time-to-signal?  Ideally I'd like to get rid of this job on pull request.

If we keep it why do we need so many paths enumerated, that seems fragile. I'd want larger globs

.github/workflows/sanitizers.yml
```diff
@@ -160,11 +160,33 @@ jobs:
...
             --test_tag_filters=-fuzz_target \
             --build_tag_filters=-fuzz_target \
             //donner/...
+
+      - name: Replay untrusted-input fuzzer corpora with UBSan
+        run: |
+          bazelisk test \
+            --config=ubsan \
+            --test_output=errors \
+            //donner/base:path_fuzzer \
```
jwmcglynn, line 177: Similar commetns to the above, I don't want targets enumerated in yml and want labels when possible

donner/base/fonts/tests/Woff2Parser_fuzzer.cc
```diff
@@ -1,9 +1,152 @@
...
+    std::abort();
+  }
+}
+
+}  // namespace
 
 /// Fuzzer entry point, see https://llvm.org/docs/LibFuzzer.html
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  // Keep the regression seed under libFuzzer's per-input watchdog even when mutation starts empty.
+  ReplayInvalidSignatureRegressionSeed();
```
jwmcglynn, line 144: Why can't these be in the corpus instead?

PR #945, opened 2026-08-16
perf: Record Geode filter passes in the shared frame encoder and pool filter textures

jwmcglynn:
Records every Geode filter pass (source rendering, compute passes, viewport copies, compositing) into the frame's shared command encoder, cutting a small blur's submissions from 11 to 2, and pools filter intermediate textures by exact format and extent so an unchanged frame creates zero filter textures. A transparent-clear pass initializes pooled outputs so stale pixels cannot leak through partial writes. Measured on an Arc A380: the Splash settled render improves 25 percent with the shared encoder and 4.5 percent more with pooling, with byte-identical output. The shared encoder chunks its command buffer every 64 filter passes, bounding pathological graphs (a 9,999-pixel-radius morphology decomposes into 1,614 passes) while small filters keep the two-submission shape.

Verification: focused filter, golden, and perf tests plus the full Geode repository gate at this layer's head, all green. The new chunking regression asserts a bounded submission range for a huge-radius morphology (10 chunk submits plus frame and readback at this fixture's fixed canvas) and at most 2 submissions for a small blur.

Post-review hardening from the stack code review: chunk boundaries on hardware Vulkan now wait for submitted work before recording continues, restoring the cross-submit storage-write to sampled-read guarantee that the deleted per-pass workaround provided; the 64-pass chunk counter is frame-scoped on the filter engine rather than per filter graph, so a document with many filtered elements cannot grow one unbounded command buffer; the shared texture pool keys on mip level count, sample count, and dimension in addition to size, format, and usage; and code comments no longer reference design docs or milestones.


donner/svg/renderer/RendererGeode.cc
```diff
@@ -1061,35 +1061,16 @@ struct RendererGeode::Impl : public geode::GeometryDebugSink {
   std::optional<RendererGeodeTextureSnapshot> borrowedTargetSnapshot;
 
   // Single CommandEncoder owned by RendererGeode for the whole frame
-  // (design doc 0030 Milestone 3). All `GeoEncoder` instances created
+  // (design doc 0030 Milestone 3). All GeoEncoder instances created
```
jwmcglynn: Do not refer to any design docs/milestones in code comments, those are banned. This is preexisting, but remove it anyway when you see it

jwmcglynn: 🤖 Removed in e2ea16ad, which rewrites this comment block without the design-doc/milestone references. The rule is now durable: a repo-wide sweep of the remaining references lands as a stacked PR on top of this stack, together with CLAUDE.md/AGENTS.md sections banning design-doc and planning-label references in code comments (including Donner's own docs/design_docs).

PR #946, opened 2026-08-16
perf: Add cross-engine benchmark with TinySkia mode and fix band boundary coverage

jwmcglynn:
Adds engine_compare_bench, a cross-engine SVG performance probe that separates device setup, parse, first settled frame, and unchanged second frame with allocation and RSS telemetry, plus a --backend=tiny-skia mode that runs Donner's in-tree CPU renderer through the same passes for a direct GPU-vs-CPU comparison. Also fixes a two-pixel Sunburst hole where a curve ending exactly on a band boundary was omitted from the following cell, and moves bounded band-selection scratch to the stack.

Verification: focused encoder, perf, and golden tests plus the full Geode repository gate at this layer's head, all green. The benchmark links both backends directly like the parity test binary, which the geode_excludes_tiny_skia_audit permits.

Post-review hardening: the CPU band binning is padded by a quotient-space epsilon in both directions so relaxed-precision GPU division cannot re-open the boundary hole on near-boundary endpoints (extra references are cost-only; the shader re-filters by exact axis ownership), the band-count bound the fixed-size scratch arrays rely on is release-asserted, and the benchmark now reports a per-run pixel fingerprint (FNV-1a plus nonzero byte count) so a stale or empty frame cannot masquerade as a speedup, measures RSS on macOS via task_info with an explicit rss_supported flag, and labels ALLOC lines cpp_heap_only because wgpu-native's Rust-side allocations are invisible to the operator-new tracker.

### DIFF
--- a/docs/design_docs/0030-geode_performance.md
+++ b/docs/design_docs/0030-geode_performance.md
@@ -429,6 +429,23 @@ algorithm.
         `drawCalls` so regressions in detection vs. batching can
         be triaged independently.
 - [ ] Milestone 7: Opportunistic cleanups.
+  - [x] Keep band-selection and band-fill scratch storage on the stack.
+    `kMaxBands` already bounds these arrays at 256 entries, so the
+    encoder now initializes only the active prefix and reuses the
+    count array as its write cursor. On allocation-instrumented first
+    frames this removes 6,010 allocation calls for Ghostscript Tiger
+    (55,914 -> 49,904, -10.8%) and 2,194 for Lion
+    (14,885 -> 12,691, -14.7%) without a measurable encode-time
+    regression.
+  - [x] Keep curve maxima inclusive when assigning band references. An
+    exclusive maximum dropped a curve from the following cell when its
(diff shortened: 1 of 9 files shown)


donner/svg/renderer/RendererGeode.cc
```diff
@@ -1061,11 +1061,14 @@ struct RendererGeode::Impl : public geode::GeometryDebugSink, public geode::Filt
   std::optional<RendererGeodeTextureSnapshot> borrowedTargetSnapshot;
 
   // Single CommandEncoder owned by RendererGeode for the whole frame
-  // (design doc 0030 Milestone 3). All `GeoEncoder` instances created
+  // (design doc 0030 Milestone 3). All GeoEncoder instances created
```
jwmcglynn: Design doc numbers are milestones are banned in code comments - remove

jwmcglynn: 🤖 Fixed in the parent layer's rewrite (e2ea16ad on codex/geode-filter-frame-encoder); after the restack this PR no longer touches that comment.

PR #947, opened 2026-08-16
perf: Move snapshot unpremultiplication to the GPU and cut map-wait polling latency

jwmcglynn:
Snapshot readback now runs a compute pass (snapshot_unpremultiply) that converts the premultiplied render target to straight-alpha RGBA8 in a pooled staging texture, so the CPU only strips row padding; the shader replicates the CPU round-half-up formula byte-for-byte and a usage-flag-forced parity test pins byte equality against the CPU path. The native map wait polls with a 100 us sleep instead of 1 ms, cutting the latency floor without changing cancellation semantics. RendererBench gains retained second-draw, snapshot, and settled phases plus readback poll stats.

Verification: parity, perf, interruptible-snapshot, golden, and resvg corpus tests plus the full Geode repository gate at this layer's head, all green. The editor package size test budgets are retuned with measured evidence: js 170,862 raw / 48,316 gzip at this layer (main measured 48,284 gzip, 16 bytes under the old budget), and the retuned budget keeps about 1 percent headroom over the final stack measurement of 48,319 gzip.

Post-review hardening: the per-size snapshot readback pool is bounded with least-recently-used eviction (a window resize previously retained one staging texture and readback buffer per distinct canvas size for the device lifetime), the lazily built readback pipeline constructs under call_once per the documented cross-thread snapshot contract, a readback that exhausts the 10 second deadline reports TimedOut and skips the CPU fallback instead of starting a second full-deadline wait, the GPU path additionally gates on the texture's actual format, both mapped-range reads are null-checked and share one row-alignment helper with the pool sizing, the unpremultiply dispatch clamps to both texture sizes, and the js raw budget is retuned to match the gzip retune's headroom.

### DIFF
--- a/donner/svg/renderer/RendererGeode.cc
+++ b/donner/svg/renderer/RendererGeode.cc
@@ -4859,10 +4859,13 @@ ReadbackMapStatus MapAndWaitReadback(const std::shared_ptr<geode::GeodeDevice>& 
 #endif
     // Never ask wgpu-native to block until the GPU map completes: a low-priority thumbnail must
     // observe a superseding main-document request promptly. Native polling is
-    // non-blocking and gets a short sleep to avoid spinning.
+    // non-blocking and gets a short sleep to avoid spinning. The 100 us sleep
+    // bounds the snapshot latency floor without the 1 ms quanta that used to
+    // dominate small-fixture readbacks; the poll itself processes the map
+    // completion callback as soon as the GPU delivers it.
     device->pollSuspending(false);
 #ifndef __EMSCRIPTEN__
-    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    std::this_thread::sleep_for(std::chrono::microseconds(100));
 #endif
   }
   device->recordReadback(usedTimedWaitAny, pollIter);
--- a/donner/svg/renderer/RendererGeode.cc
+++ b/donner/svg/renderer/RendererGeode.cc
@@ -294,10 +294,12 @@ double computeBlurPadding(const components::FilterGraph& filterGraph) {
 }
 
 /// WebGPU requires bytesPerRow alignment to 256 when copying textures to
-/// buffers. This rounds the unpadded row width up to the next 256 boundary.
+/// buffers. Delegates to the shared helper so the renderer's map-range math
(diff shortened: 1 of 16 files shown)


donner/svg/renderer/geode/GeodeDevice.h
```diff
@@ -24,6 +24,30 @@ class GeodeGradientPipeline;
...
 class GeodeMaskPipeline;
 class GeodeFilterEngine;
+class GeodeSnapshotReadbackPipeline;
+
+/**
+ * Move-only resource set for one GPU snapshot unpremultiply readback: a
+ * straight-alpha staging texture, its view, and a map-readable readback
+ * buffer. Acquired from and returned to the per-device pool keyed by size,
+ * so repeat snapshots at the same dimensions allocate nothing (design doc
+ * 0030 Milestone 4.2 pooling).
```
jwmcglynn: Remove design doc and milestone reference - those are banned

donner/svg/renderer/RendererGeode.cc
```diff
@@ -1061,11 +1062,14 @@ struct RendererGeode::Impl : public geode::GeometryDebugSink, public geode::Filt
   std::optional<RendererGeodeTextureSnapshot> borrowedTargetSnapshot;
 
   // Single CommandEncoder owned by RendererGeode for the whole frame
-  // (design doc 0030 Milestone 3). All `GeoEncoder` instances created
+  // (design doc 0030 Milestone 3). All GeoEncoder instances created
```
jwmcglynn: Remove design doc ref

donner/svg/renderer/geode/GeodeDevice.h
jwmcglynn: 🤖 Removed in f150beb8; the comment now documents the pool's bounded LRU semantics instead.

donner/svg/renderer/RendererGeode.cc
jwmcglynn: 🤖 Removed: the frameCommandEncoder block was rewritten at the bottom layer (e2ea16ad) and the readback pooling comment in this layer was cleaned in f150beb8.

PR #948, opened 2026-08-16
perf: Add GPU residence for gradient fills and reduce filter resource churn

jwmcglynn:
Gradient-painted fills move off the per-frame arena path: GeodeResidentGradientSlot persists each entity's encoded geometry and 672-byte gradient uniform in a resident buffer with a cached 11-binding bind group, and only rewrites the uniform when it changes. Filter primitives bump-allocate uniforms from one per-frame scratch buffer, the texture pool cap rises from 8 to 16 entries per exact size, blur passes alternate two ping-pong intermediates, and the per-primitive subregion clip is folded into the final blur pass for axis-aligned transforms. Zero-deviation blurs keep the dedicated clip pass, pinned by a regression test that fails before and passes after.

Verification: goldens unchanged, and the full Geode repository gate at this layer's head is green.

Post-review hardening: the folded blur clip engages only when every rounded corner fits in int32 (a pathological CTM previously made the double to int32 conversion undefined behavior, observed collapsing to INT32_MIN and blanking the blur output), the per-frame filter engine reset is documented accurately (pass bind groups are created per pass; pooled texture identities rotate across frames), and a pixel regression test pins that a gradient stop edit invalidates the resident gradient draw while an unchanged document renders identically.


donner/svg/renderer/geode/shaders/gaussian_blur.wgsl
```diff
@@ -19,10 +19,23 @@ struct BlurParams {
   kernel_type: u32,// 0 = Gaussian, 1 = Box
   box_left: i32,   // box mode: number of samples to the negative side
   box_right: i32,  // box mode: number of samples to the positive side
-  _pad0: u32,
+  clip_min: vec2i, // optional output clip (see clip_active)
+  clip_max: vec2i, // optional output clip: zero pixels with coord < min or >= max
+  clip_active: u32,
```
jwmcglynn, line 31: Add doc comment

donner/svg/renderer/geode/GeoEncoder.h
```diff
@@ -561,6 +562,35 @@ class GeoEncoder {
   void fillPathRadialGradient(const Path& path, const RadialGradientParams& params, FillRule rule,
                               const EncodedPath* precomputedEncoded = nullptr);
 
+  /**
+   * Fill a path with a linear gradient through a persistent GPU residence
+   * slot (design doc 0030 wave 2 extension for gradient paints).
```
jwmcglynn: Remove design doc refs

donner/svg/renderer/geode/shaders/gaussian_blur.wgsl
jwmcglynn: 🤖 Documented in ca82a138: the clip rectangle fields, clip_active semantics, and the explicit tail padding now carry doc comments.

donner/svg/renderer/geode/GeoEncoder.h
jwmcglynn: 🤖 Removed in ca82a138; the resident-gradient docs now describe the fallback conditions inline without planning labels.

PR #1015, opened 2026-08-18
Preserve WebGPU map cleanup across handle reuse

jwmcglynn:
This fixes a browser WebGPU buffer-map cleanup race that can terminate the renderer worker after a cancelled snapshot is followed by a new readback.

- give each asynchronous Emdawn buffer map a distinct cleanup-list identity
- delete rejection cleanup state only when that map generation is still current
- add a source contract that fails if cleanup returns to raw handle-only deletion
- preserve prompt cancellation without retries, longer timeouts, or weakened browser assertions

## Root cause

Cancelling a pending readback completes the C++ map future synchronously, so the released buffer object's address can be reused before the original JavaScript Promise rejects. The old rejection used that address as an unguarded table key and could delete the new map's cleanup list. The next readback then mapped successfully but crashed while registering its cleanup callback.

The fix captures the list installed by each map and removes the table entry only when it still points to that exact list. Late rejection can no longer mutate a newer generation at the same handle address.

## Validation

- regression contract: red on the old bridge, green after the generation guard
- `npm --prefix donner/editor/wasm/tests run test:selector` (28 passed)
- both Emdawn Wasm link consumers and the editor package-size test
- hermetic Bazel Chromium smoke test
- formerly failing Text and Style browser case, 20 of 20 headed Chromium runs
- full browser matrix: 26 Chromium default, 5 Firefox resize, 2 WebKit carousel, 9 Firefox composited, and 13 Chromium composited tests passed
- `bazel test //... --test_output=errors` (458 targets: 341 passed, 117 intentional skips)
- repository lint, CMake validation, dprint, and diff checks

### DIFF
--- a/donner/editor/wasm/tests/wgpu-readback-contract.spec.mjs
+++ b/donner/editor/wasm/tests/wgpu-readback-contract.spec.mjs
@@ -31,6 +31,52 @@ const geodeDeviceHeader = await readFile(
   new URL("../../../svg/renderer/geode/GeodeDevice.h", import.meta.url),
   "utf8",
 );
+const emdawnWebgpuSource = await readFile(
+  new URL("../../../../third_party/emdawnwebgpu/webgpu/src/library_webgpu.js", import.meta.url),
+  "utf8",
+);
+
+test("late map rejection cannot clear a reused buffer handle's cleanup state", () => {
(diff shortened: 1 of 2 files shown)


jwmcglynn:
@copilot  rebase this PR

PR #1017, opened 2026-08-18
feat: Add a composited rendering mode setting to the editor

jwmcglynn:
Adds a View menu "Composited Rendering" setting with three runtime modes, while preserving the editor's paint-order layer presentation in every non-Off mode.

## Modes

| Mode | Behavior |
| --- | --- |
| On (default) | Full compositing. Mandatory isolation, editor selection/drag layers, animation promotion, and complexity bucketing are enabled. |
| Filters Only | Only SVG filters retain automatic content-isolation layers. Opacity groups, blend modes, and masks render inline; editor selection/drag layers remain active; animation promotion and complexity bucketing are disabled. |
| Off | No `CompositorController` is constructed. Each frame renders directly and is presented as the explicit `full-canvas` diagnostic path. |

All three modes produce identical final pixels. They trade retained caching and layer maintenance against per-frame rendering work.

## Active presentation contract

Normal active presentation in On and Filters Only always publishes the compositor's paint-order tiles. It never falls through to `full-canvas` because the editor is idle, a selection is not being dragged, the first-frame cache is cold, or an interaction split is unsafe.

- A cheap selected scene presents background, selected, and foreground as three immediate-classified layers.
- Filters Only publishes the splash's three blurred glow groups as separate filter layers.
- Selection and drag layers remain structural editor behavior in both non-Off modes.
- Overview infill is an out-of-band low-resolution coverage cache. Geometry debug is a separate explicit flat diagnostic named `geometry-debug-flat`.

## Promotion safety

`PromoteResult::OwningTilesRequired` replaces the old full-canvas fallback contract. If extracting a requested selection would split an ancestor compositing context or overlap an assigned ancestor or descendant, the entity remains inside the paint-order tiles that own it.

Detector and interaction reconciliation now run against the prepared render tree. This covers cold requests, filter additions, `display:none`, opacity/isolation changes, structural remaps, and stale empty hint components in the same accepted frame.

`CompositedPreview::entity` names a genuinely movable promoted tile only. Off, owning-tile refusal, and partially covered nested multi-selection publish a null movable entity, so the scheduler requests rendered drag frames instead of moving chrome over stationary pixels. Nested parent/child selection is overlap-safe in both ordering directions.

## Verification

- Full repository gate: 342 passed, 117 intentional skips, 0 failures.
- Uncached native and Geode async-renderer/compositor suites.
- Editor presentation scheduler and editor-control MCP session suites.
- Editor Wasm transition and package-size gates.
- Hermetic Chromium remote smoke test.
- Repository lint, changed-line clang-format, and diff checks.
- Independent code review and public privacy review passed on the final candidate.

### DIFF
--- a/donner/editor/AsyncRenderer.cc
+++ b/donner/editor/AsyncRenderer.cc
@@ -9,6 +9,7 @@
 
 #include "donner/base/MemoryAttribution.h"
 #include "donner/base/Utils.h"
+#include "donner/base/xml/components/TreeComponent.h"
 #include "donner/editor/OverlayRenderer.h"
 #include "donner/editor/TracyWrapper.h"
 #include "donner/svg/SVGDocument.h"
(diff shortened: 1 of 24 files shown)


jwmcglynn:
@copilot resolve the merge conflicts on this branch.

PR #1018, opened 2026-08-18
perf: Batch clipped, stroked, and gradient-filled runs through pathID bindings

jwmcglynn:
Stacked on the font-identity PR; review that one first. This PR's own diff is
the nine pathID commits.

## Summary

Three changes let ordered batching hold runs it previously had to break:

1. **Clip rect as record data.** The scissor is opened per batched draw from a
   per-instance rect, so a clip no longer forces a solo draw.
2. **Strokes join the batch.** Stroke draws append when their encode compares
   equal, instead of always starting a new run.
3. **Gradient fills resolve by record index.** A 25-row paint block lives in the
   residence slot at binding 11, and the batched shader entry point reads the
   gradient through the record rather than a per-draw uniform.

## Correctness work, which is most of the diff

Batching a gradient made a latent record-lifetime hazard reachable, and the
fixes are worth reviewing more closely than the feature itself.

**Paint replay from the append, not the slot.** A batched instance captures its
paint and clip state by value when it is appended. Previously the flush re-read
those scalars from the residence slot, so a solo draw of the same entity landing
between append and flush could republish the slot and make an already-appended
gradient render as opaque black.

**Resident slots are claimed across every open frame generation.** The slot
reuse gates compared the stamp for equality against the current frame. Frame
generations are per device, so an offscreen renderer created from the same
device mid-frame (an filter fragment, a layer thumbnail) gets a new generation,
reads "slot free", and can rewrite a gradient paint block or record that the
outer frame's already-recorded draws still reference; the queued write lands
before the outer submit. The gates now ask whether the stamp falls within any
open generation, with the never-drawn sentinel handled explicitly. Five gates
were converted, including the two resident gradient paths.

**Stable buffer ids through every batch opener.** The glyph and gradient batch
openers left the bind-group cache keys unstamped, which silently disabled
caching for those batches rather than failing.

## Tests

- A shared-device frame test covering the stamp predicate directly, plus an
  end-to-end case that records a gradient batch, mutates the gradient, renders
  an offscreen pass from the same device mid-frame, and asserts the outer frame
  keeps its original pixels. It fails without the guard.
- Gradient ordering cases for solid-then-gradient, gradient-then-solid,
  gradient-twice, and mixed runs of distinct entities on one slot.
- Counter ratchets pinned per configuration.

## Counter results, verified at this head

Draw and buffer-write counts are deterministic and pinned by the corpus tests,
so these are current rather than historical:

- `z0rly_test6` previously needed a bind-group-create ceiling of 22 and now
  reaches the shared target outright, so its entry is deleted. Its own tracking
  note predicted exactly this once bind groups were cached across draws sharing
  a pipeline.
- Two `use`-instance scenes reach zero buffer writes with batching while still
  needing their ceiling without it.

To express that honestly the ratchet gained a build-aware violated mask
alongside the existing ceiling, because batching can clear a violation outright
rather than only lower its number, and an entry still naming a counter the
shipped build already holds would hide the next regression behind a stale
ceiling.

Wall-clock figures measured earlier on this branch are deliberately not quoted
here: the correctness fixes above landed after that run, and a re-measurement at
this exact head is in progress. I will post it as a comment rather than claim a
number this head has not produced.

## Verification

Both compile-time batching states pass on discrete-GPU hardware: 81 pass, 3
config-gated skips. The gradient perf cases and both shared-device frame cases
are green by name.

Known, inherited from the font/text PRs below this one and not introduced here:
with batching compiled off, the counter corpus is red on three text scenes,
because the ratchet rework in that PR made the entries describe the shipped
build only. Verified by running that branch alone with the flag off and getting
the identical three failures. The compile-time constant means CI never exercises
that configuration; it affects only manual bisection builds.

### DIFF
--- a/donner/svg/renderer/RendererGeode.cc
+++ b/donner/svg/renderer/RendererGeode.cc
@@ -1032,12 +1032,14 @@ std::shared_ptr<RendererGeodeTexturePool> TexturePoolForDevice(geode::GeodeDevic
 
 }  // namespace
 
-/// Gate for ordered cross-entity batching. Enabled: the deferred flush that
-/// used to reorder a batched draw past a scissor or clip change can no longer
-/// form one, because the eligibility predicate refuses any draw taken while a
-/// clip state, a scissor, or a mask pass is active. The same-entity instanced
-/// path and the solo residence flows are unchanged and still handle
-/// everything a scene batch declines.
+/// Gate for ordered cross-entity batching. Enabled: a batch's single draw can
+/// no longer observe target state that moved on after its instances were
+/// appended. A clip rectangle travels in each instance's record and the
+/// fragment stage applies it, and the eligibility predicate still refuses any
+/// draw taken under a clip polygon, a clip mask, or an open mask pass, which
+/// are the states a deferred draw would otherwise read at flush time. The
+/// same-entity instanced path and the solo residence flows are unchanged and
+/// still handle everything a scene batch declines.
 ///
 /// Kept as a compile-time constant rather than a runtime option so that
 /// turning it off folds the whole eligibility predicate away, leaving the
@@ -2025,11 +2027,11 @@ struct RendererGeode::Impl : public geode::GeometryDebugSink, public geode::Filt
     /// Encoder clip-state version at the batch's first append. A batch
     /// never spans a clip change (one draw carries one clip state).
     uint64_t sceneClipVersion = 0;
-    /// GPU-residence slot for this source entity's fill (captured
-    /// when the batch starts). Used only by the size-1 flush path - a
-    /// size >= 2 flush routes through the instanced arena path, which has
-    /// no per-entity residence.
-    geode::GeodeResidentSlot* residentFillSlot = nullptr;
+    /// GPU-residence slot for the encode this batch started on - the source
+    /// entity's fill slot for a fill, its stroke slot for a stroked outline.
+    /// Used only by the size-1 flush path; a size >= 2 flush routes through
+    /// the instanced arena path, which has no per-entity residence.
+    geode::GeodeResidentSlot* residentSlot = nullptr;
   };
   std::optional<PendingBatch> pendingBatch;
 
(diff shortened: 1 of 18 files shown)


jwmcglynn:
## Wall-clock re-measurement at this head

As promised in the description, here are numbers measured at `7691855a` rather
than carried over from the earlier run.

Method: discrete GPU, `-c opt`, shipped batching-on build. Baseline is the
branch directly below this PR, so the delta isolates this PR's work rather than
including the text and font wins. Both binaries were built first, then run
alternating baseline/head, 5 runs per side, so host drift hits both sides
equally. Each figure is the median of the 5 per-run medians; the bracket is the
full range of those medians.

| scene | phase | baseline | head | delta |
| --- | --- | --- | --- | --- |
| Tiger | steady | 4.175 ms [3.344-4.405] | **1.600 ms** [1.544-1.723] | **-61.7%** |
| Tiger | mutating | 7.972 ms [7.409-8.383] | 5.595 ms [5.522-5.772] | -29.8% |
| Tiger | first draw | 42.072 ms [37.594-42.653] | 35.823 ms [34.140-38.127] | -14.9% |
| gradient_grid | steady | 3.404 ms [3.193-3.610] | **0.805 ms** [0.737-0.919] | **-76.4%** |
| gradient_grid | mutating | 5.457 ms [5.312-5.852] | 3.061 ms [3.037-3.213] | -43.9% |
| gradient_grid | first draw | 12.860 ms [12.127-13.609] | 8.507 ms [8.324-9.150] | -33.8% |
| lion | first draw | 9.673 ms [9.082-10.266] | 8.427 ms [7.972-9.012] | -12.9% |

Steady-frame draw calls, which are deterministic and pinned by the corpus
tests: gradient_grid 96 to 1, Tiger 153 to 16, complex_scene 16 to 11,
moderate_paths 6 to 5, lion and simple_shapes unchanged at 2.

### What is deliberately not claimed

`complex_scene`, `moderate_paths`, and `lion` steady all land inside their own
run-to-run spread in both directions, so their small improvements are not
claimed even though their draw-call collapse is real and deterministic.
`simple_shapes` first draw shows +7.2%, which is well inside that scene's
12-22% spread on a 0.75 ms measurement; it is noise rather than a regression.
Nothing measured regressed outside spread.

### Effect of the correctness fixes

Tiger steady is 1.600 ms here against 1.64 ms in the pre-fix run, which is
indistinguishable inside an 11% spread. The frame-generation guard, buffer-id
stamping, and bind-group LRU therefore cost no measurable steady-state time, as
expected: the guard widens a comparison already on that path, the ids are
compared inside an existing tuple comparator, and the LRU refresh runs only on a
cache hit over at most 256 small keys.

The headline percentage moved from the earlier -58.4% only because the baseline
moved; the head's absolute time did not.

PR #1029, opened 2026-08-21
Harden CI dependency and release provenance

jwmcglynn:
Pins CI dependencies and makes release artifacts a single-attempt, digest-verified handoff from reviewed source.

- uses released GitHub Action and Bazel dependency tags so Renovate follows published releases
- verifies downloaded executable archives by SHA-256 while keeping generated Bzlmod lock state local
- removes write-capable BuildBuddy credentials from PR and arbitrary-branch execution paths
- builds Linux and macOS binaries once, compares both provenance manifests to the tagged source, attests the binaries, and uploads without replacement
- keeps BCR publication separate and operator-initiated after release artifact verification
- routes every deterministic fuzzer corpus into the UBSan sweep and derives text-full variant coverage automatically

## Validation

- `bazel test //...`: 350 passed, 120 configuration-intentional skips, 0 failures
- `tools/lint.sh`: 2,588 files clean
- `python3 tools/cmake/gen_cmakelists.py --check`: passed
- CI workflow policy and CMake generator tests: passed
- dependency tags, archive digests, release permissions, and build provenance independently reviewed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,16 +1,30 @@
 name: Release
 
-# Trigger on when a release is created or changed.
+# Build and publish an immutable artifact set exactly once for each release.
 on:
   release:
-    types: [published, edited]
-  workflow_dispatch:
+    types: [published]
+
+permissions:
+  contents: read
 
 jobs:
   linux:
+    if: github.run_attempt == 1
     runs-on: ubuntu-24.04
+    env:
+      RELEASE_TAG: ${{ github.event.release.tag_name }}
     steps:
       - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+          ref: ${{ github.event.release.tag_name }}
+
+      - name: Verify release source
+        run: |
+          test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
+          test "$(git rev-list -n 1 "refs/tags/${RELEASE_TAG}")" = "$GITHUB_SHA"
 
       - name: Install Clang dependencies
         run: |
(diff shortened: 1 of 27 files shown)


.github/workflows/release.yml
```diff
@@ -55,45 +88,98 @@ jobs:
...
+            release/linux/donner-svg_linux_x86_64.sha256
+            release/linux/donner-svg_linux_x86_64.provenance
+            release/macos/donner-svg_darwin_arm64
+            release/macos/donner-svg_darwin_arm64.sha256
+            release/macos/donner-svg_darwin_arm64.provenance
+          )
+          for asset in "${assets[@]}"; do
+            name="${asset##*/}"
+            gh api --method POST -H "Content-Type: application/octet-stream" \
+              "${upload_base}?name=${name}" --input "$asset"
```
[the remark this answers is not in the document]
jwmcglynn: Fixed in 6a6426a1. Release publication now uses an ID-bound, digest-aware publisher with three bounded attempts. It resumes partial publication only when an existing asset has the exact local SHA-256, rejects mismatches, and causally tests matching, mismatched, absent, lost-success, and exhaustion states.

.github/workflows/release.yml
```diff
@@ -55,45 +88,98 @@ jobs:
...
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: |
+            release/linux/donner-svg_linux_x86_64
+            release/macos/donner-svg_darwin_arm64
+
+      - name: Reverify release tag before publication
+        run: |
+          test "$(git rev-parse HEAD)" = "$GITHUB_SHA"
+          test "$(git rev-list -n 1 "refs/tags/${RELEASE_TAG}")" = "$GITHUB_SHA"
```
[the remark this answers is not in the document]
jwmcglynn: Fixed in 6a6426a1. The publisher runs git ls-remote immediately before uploads, prefers the peeled commit for annotated tags, and requires it to equal GITHUB_SHA. Lightweight, annotated, missing, and mismatched cases are covered by executable tests.

third_party/bazel/non_bcr_deps.bzl
```diff
@@ -55,8 +55,8 @@ def _non_bcr_deps_impl(_mctx):
     new_git_repository(
         name = "harfbuzz",
         build_file = "//third_party:BUILD.harfbuzz",
+        commit = "56feae4035bdd48f62ba2b8d8c16232d4d89b3a4",
         remote = "https://github.com/harfbuzz/harfbuzz.git",
-        tag = "14.2.1",
```
jwmcglynn, line 59: For bazel, I do not want to pin a commit, as it conflicts with renovate infra - with commits renovate will try to upgrade whenever there is a new commit, not just a new tag.

This does change our security stance slightly, but it provides us more security in other aspects as we don't need to handle frequent updates.

.github/workflows/sanitizers.yml
```diff
@@ -124,7 +129,9 @@ jobs:
     timeout-minutes: 120
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```
jwmcglynn: Same comment for GH workflows, we should use @v7 so renovate doesn't try to upgrade to unreleased versions

MODULE.bazel.lock
jwmcglynn, line 1: Do not check in .bazel.lock files

third_party/bazel/non_bcr_deps.bzl
jwmcglynn: Updated in c30d18f2. HarfBuzz is back on tag 14.2.1 and Tracy on v0.13.1, so Renovate tracks published releases rather than arbitrary commits. The policy test now permits released semver tags, with explicit commits only for repositories without a release-tag path.

.github/workflows/sanitizers.yml
jwmcglynn: Updated in c30d18f2. Every external GitHub Action now uses its released version tag, for example actions/checkout@v7. The workflow policy rejects branch names and inconsistent versions while remaining Renovate-compatible.

MODULE.bazel.lock
jwmcglynn: Updated in c30d18f2. MODULE.bazel.lock is absent from both commits and restored to .gitignore. Release builds explicitly use --lockfile_mode=off and compare provenance over the versioned module and configuration files instead.

PR #1099, opened 2026-09-08
Keep canvas shortcuts out of active Inspector inputs

jwmcglynn:
Command/Ctrl+A while editing an Inspector value selected canvas elements instead of the field's text. Keep canvas editing shortcuts from consuming keys owned by text inputs, while preserving file shortcuts.

Add coverage through the shortcut handler and an editor-control replay that changes X from 32 to 64 and checks the resulting document geometry. Extract file-shortcut handling into a small helper to keep the routing readable.

Validation: the full repository test gate passes (479 targets passed; 30 platform/configuration skips), including the EditorShell and editor-control suites. The same recorded interaction fails before the fix and passes afterward. CMake generator validation, formatting, and scoped lint pass.

### DIFF
--- a/donner/editor/EditorShell.cc
+++ b/donner/editor/EditorShell.cc
@@ -2603,6 +2603,36 @@ void EditorShell::applyMenuActions(const MenuBarActions& menuActions) {
   }
 }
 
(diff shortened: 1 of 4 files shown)


jwmcglynn:
Approved

PR #1100, opened 2026-09-08
Compile only the shader components each target needs

jwmcglynn:
Separate the typed shader IR, individual emitters, and Metal binding-map header into precise build targets. Metal no longer compiles the eight shader translation units merely to consume binding constants, and the WGSL generator depends only on its emitter.

Move RecordingDevice into a test-only library and update its test consumers. The shared validation-tool gate remains intact.

Validation: full repository test gate passes (479 targets passed; 30 platform/configuration skips), CMake generator validation passes, and formatting is clean. These dependency-closure reductions do not imply a measured linked-binary-size reduction.


jwmcglynn:
Approved

PR #1101, opened 2026-09-08
Preserve native GPU upload and resource ordering

jwmcglynn:
Preserve native GPU operation ordering across uploads, dispatches, and rendering. Metal queues bounded upload batches so an earlier submission cannot observe a later host write. Vulkan transitions images at their actual uses and includes attachment loads and vertex texture reads in the required access scopes.

Coalesce repeated Metal writes, discard unsubmitted writes as soon as ordinary buffer or texture handles retire, and enforce the configured budget against logical payload bytes while separately bounding packed staging memory. Synchronize only GPU-written managed buffers. Remove incompatible aggregate render-pass dependencies and make Vulkan barrier history opt-in; the Vulkan production implementation loses 167 lines.

Add deterministic in-flight upload and dispatch-transition regressions, including unused/replaced bindings, repeated writes, explicit/RAII retirement, and padded texture rows. The initial full repository gate passed 480 targets with 30 platform/configuration skips. The retirement and budget follow-up passes 256 shared GPU and native Metal cases with no skips under Metal validation. CMake generation, formatting, and changed-method lint pass; native Vulkan comparisons were separately checked with synchronization validation enabled. Broader GPU hardening continues separately.

### DIFF
--- a/donner/gpu/Device.cc
+++ b/donner/gpu/Device.cc
@@ -423,7 +423,16 @@ void Device::recycleRetiredSlot(ResourceKind kind, uint32_t slotIndex) {
   }
 }
 
+void Device::onRetireBuffer(uint32_t) {}
+
+void Device::onRetireTexture(uint32_t) {}
+
 void Device::retireResource(ResourceKind kind, uint32_t slotIndex, uint64_t lastUseSerial) {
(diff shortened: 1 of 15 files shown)


jwmcglynn:
Approved

PR #1122, opened 2026-09-09
Preserve filter parameters across shared renderer frames

jwmcglynn:
Preserve shared filter parameters while another renderer still has an open frame. Starting a blue filtered frame previously overwrote the parameters of an unsubmitted red frame, causing both frames to render blue regardless of submission order.

Close the current generation before checking for other open frames, and reset shared scratch only when no frame remains open. Generation cleanup now belongs to implementation teardown, so move-assigning over an open renderer cannot leave a phantom generation that prevents later scratch reuse.

Regression tests first reproduced both color corruption and renewed buffer growth after move-assignment. They cover both submission orders and verify zero new buffers after warm-up across settled filter frames, including replacement of an open renderer.

Indefinitely overlapping frames still require further working-set improvements; this change does not raise memory limits.

Validation: full Bazel suite (488 targets passed; 31 configuration-incompatible targets skipped), CMake generator, lint, formatting, GPU inventory, and independent code/security review.

### DIFF
--- a/donner/svg/renderer/RendererGeode.cc
+++ b/donner/svg/renderer/RendererGeode.cc
@@ -4452,13 +4452,14 @@ struct RendererGeode::Impl : public geode::GeometryDebugSink,
     if (frameResourceScopeDepth == 0) {
       resetOwnedFrameBudgets();
     }
-    if (device) {
+    closeFrameGeneration();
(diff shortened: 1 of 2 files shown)


jwmcglynn:
@codex review

PR #1126, opened 2026-09-10
Execute the tile filter through the typed GPU runtime

jwmcglynn:
Express feTile as one typed shader program and record it through the shared GPU command stream. Build-time WGSL generation replaces the handwritten shader and raw WebGPU factory; native Metal and Vulkan execute the same program through their respective emitters.

The migration preserves signed-origin wrapping, transparent output for degenerate rectangles, float intermediates, and allocation-refusal propagation. Source samples use the source texture's bounds. Tests cover odd-sized dispatches, differing input/output extents, fractional float values, signed and above-one storage values, generated artifact identity, and existing renderer pixels and budgets.

The Geode editor window test now uses its existing runfiles-aware variant launcher in CI. Its direct executable could fail before entering tests on a clean worker because the dynamic loader could not locate a declared shared library.

Validation: 491 full-suite targets passed with 33 configuration skips; all 13 owning shader/renderer/editor targets and five native Vulkan targets passed. Direct Metal execution, shader compiler validation, CMake generation, lint, and inventory checks passed. Pixel tolerances and memory caps are unchanged.


.github/workflows/main.yml
```diff
@@ -1003,16 +1003,10 @@ jobs:
...
-      # this step they run on no lane at all. Measured on Apple Silicon with a
-      # warm build: 74 s wall, ~62 s of it test execution (`test:geode` pins
-      # `--local_test_jobs=1`, so the suite is serial). The whole
-      # `//donner/editor/...` tree under `--config=geode` is 260 tests / 193 s
-      # if this ever needs widening.
+      # Exercise the Geode editor configuration explicitly. The window test uses its
+      # transitioned wrapper to locate declared shared libraries on clean workers.
       GEODE_EDITOR_TARGETS: >-
-        //donner/editor/tests:editor_window_tests
+        //donner/editor/tests:editor_window_tests_geode
```
jwmcglynn, line 1009: File a task for a follow-up: Remove any direct-target refs in .yml files, switch to using labels if we need to filter (tech debt)

jwmcglynn: Tracked in #1129. The follow-up covers removing direct per-target references from workflow YAML while preserving lane coverage and label-based selection.

PR #1127, opened 2026-09-10
Execute bounded morphology through the typed GPU runtime

jwmcglynn:
Express feMorphology as a typed shader program and record its erosion and dilation passes through the shared GPU command stream. Build-time emitted WGSL replaces the handwritten shader and raw WebGPU factory.

The host retains the existing 31-pixel per-pass radius cap and two-texture scratch reuse for larger kernels. Allocation, texture-import, and dispatch refusal return no output before later passes can consume an incomplete result. The shader preserves the existing component-wise operations and transparent-border behavior.

Tests cover zero radius, both axes, erosion and dilation, transparent edges, fractional float storage, direct Metal/MSL and Vulkan/SPIR-V execution, and equality between generated and embedded shader artifacts. The existing radius-255 two-axis regression verifies repeated-pass scratch accounting.

Validation: 491 full-suite targets passed with 33 configuration skips; all 13 owning shader/renderer/editor targets and five native Vulkan targets passed. CMake generation, lint, and inventory checks passed. Pixel tolerances and memory caps are unchanged.

This follows #1126; the diff is limited to morphology and waits for its parent before main-target CI qualification.


donner/gpu/shader/tests/testdata/morphology.msl
```diff
@@ -0,0 +1,48 @@
+// Generated by donner::gpu::shader::MslEmitter. Do not edit.
```
jwmcglynn, line 1: These goldens are becoming unwieldy, let's stop doing this for shaders and remove existing ones. Change the design doc patterns and do a follow-up PR in the design to clean out old ones.

jwmcglynn: Addressed in #1130: removes existing generated shader snapshots and export helpers, updates the design verification pattern, and retains structural, compiler, execution, and renderer pixel checks.

PR #1308, opened 2026-09-19
Match XML parser tokenization in the export root and body scan

jwmcglynn:
Fixes the nightly fuzz assertion in `//donner/editor/tests:viewport_svg_export_roundtrip_fuzzer_soak`, where `ExportViewportAsSvg` emitted XML that its own parser rejected.

The exporter's hand-rolled source scan had drifted from the XML parser's tokenization in four ways, each reachable through markup-like text that the parser accepts but the scan misread:
- a processing instruction containing `>` followed by an `<svg>` element was cut at the first `>` instead of the PI terminator, so the embedded element was mistaken for the document root;
- an XML declaration terminator was not quote-aware, so a `?>` inside a quoted declaration value ended it early;
- a doctype internal subset was scanned without bracket or quote awareness, so a `>` inside the subset, or brackets inside a quoted entity value, derailed the doctype end; and
- the body end used the last `</svg` in the source, so a stray `</svg` sequence after the root (for example in a trailing comment) truncated the body and left the real root's `<svg>` unclosed inside the wrapper `<g>`.

In each case the export produced unbalanced or malformed XML. The scan now mirrors the parser: PI and declaration terminators, bracket-aware doctypes with quote-aware entity declarations, and a matching-close-tag search that tracks nested `<svg>` elements and skips comments, CDATA, processing instructions, and quoted attribute values.

The checked-in fuzzer corpus gains six regression seeds, including the minimized nightly finding and the additional shapes found during review.

Validation:
- Six new unit tests fail before their corresponding fixes and pass after; the existing root-scanner tests continue to pass.
- Nightly crash input on the unfixed ASan fuzzer binary: reparse assertion, abort. The same input and all six corpus seeds on the fixed ASan fuzzer binary: all execute cleanly.
- `bazel test //donner/editor/tests:viewport_svg_export_tests //donner/editor/tests:viewport_svg_export_roundtrip_fuzzer` pass, including corpus replay with the new seeds.
- `bazel test //donner/editor/tools:generate_showcase_asset_cli_tests` passes.
- `tools/lint.sh` clean; changed files are clang-format clean.
- Independent code/security review approved, including a differential parity check of the doctype scan against the XML parser across 900k randomized inputs.

Nightly run: https://github.com/jwmcglynn/donner/actions/runs/35436770701 (Linux job, `//donner/editor/tests:viewport_svg_export_roundtrip_fuzzer_soak`).


jwmcglynn:
Wrong approach; will deduplicate XML logic

jwmcglynn:
Superseding PR is #1312 (shared-tokenizer exporter, this PR's tests and corpus seeds carried over).

Issue #1212, opened 2026-09-13
P2: Restore CMake test discovery or remove the unsupported test contract


jwmcglynn:
The CMake generator discovers only the public library dependency closure, yet still emits DONNER_BUILD_TESTS and documents ctest execution. Test-only leaves cannot enter that closure.

Observed by source inspection at `9bf4e00f1f199751523e5eca307be4f239adb9ae`; no new execution results are claimed.

Acceptance criteria:

- Define the supported CMake test set and include its leaves in discovery, or remove the unsupported option and documentation together.
- If test support is retained, add a consumer-level check that enabling the documented option discovers a nonempty expected test set and executes it. If support is removed, remove the option and test-running instructions together and verify the documented consumer build.
- Keep editor-only and platform-incompatible targets outside the supported mirror.

Review evidence:

- https://github.com/jwmcglynn/donner/pull/705#discussion_r3517006650

Current source:

- [tools/cmake/gen_cmakelists.py:387](https://github.com/jwmcglynn/donner/blob/9bf4e00f1f199751523e5eca307be4f239adb9ae/tools/cmake/gen_cmakelists.py#L387): Only library root is queried
- [tools/cmake/gen_cmakelists.py:1070](https://github.com/jwmcglynn/donner/blob/9bf4e00f1f199751523e5eca307be4f239adb9ae/tools/cmake/gen_cmakelists.py#L1070): DONNER_BUILD_TESTS option still emitted
- [docs/building.md:160](https://github.com/jwmcglynn/donner/blob/9bf4e00f1f199751523e5eca307be4f239adb9ae/docs/building.md#L160): Advertises enabling and running tests

PR #1310, opened 2026-09-19
Grandfather baseline Python complexity in check_python_quality

jwmcglynn:
Grandfathers pre-existing complexity debt in the Python quality gate, matching the C++ check max(limit, baseline) semantics.

check_source() failed any touched function scoring over the limit, even when the edit reduced the score. Unlike the C++ complexity check, there was no allowance for pre-existing debt, so legitimate edits to over-limit functions could not land without a full refactor. This scores the baseline function too and fails only above both the limit and the baseline. function_score() is extracted so the gate itself stays within its own complexity budget. New functions and score increases fail exactly as before.

This unblocks the stacked #1212 generator edit, which reduces generate_all_packages from 117 to 116.

Verification: bazel test //tools:check_python_quality_tests passes; tools/lint.sh passes on this branch under the current gate; the three new tests cover reduce-pass, increase-fail, and hold-steady-pass.


jwmcglynn:
No - wrong direction. This is papering over legit tech debt. When we encounter complexity errors we must fix the offending code.

PR #1312, opened 2026-09-19
Derive viewport SVG export bounds from the shared XML tokenizer

jwmcglynn:
Replace the viewport exporter's hand-rolled XML scan with the shared tokenizer, superseding #1308.

The exporter derived root bounds, root attributes, and external-reference checks from its own source scanners, which drifted from `XMLParser` tokenization on every edge: processing-instruction terminators, quote-aware declarations, bracket-aware doctypes with entity skips, and matching close tags. PR #1308 chased parity with about 200 new scan lines; this change deletes the scanners instead. The source is tokenized once with `donner::xml::Tokenize`, and root bounds, attributes (source order, raw bytes), and the `href` check all derive from the token stream. Comments, CDATA sections, doctypes, processing instructions, and quoted values are opaque single tokens, so markup-like text inside them can never be mistaken for elements again.

Two places where `Tokenize` diverged from `XMLParser` (the `SVGParser` configuration) are converged in the shared tokenizer, with unit tests: `<?xml` terminators inside quoted values no longer end the declaration, and doctypes skip `<!ENTITY>` declarations quote-aware inside the internal subset. Highlighting and autocomplete gain the same correctness; their suites pass unchanged.

Behavior notes:
- The six PR #1308 regression tests and six corpus seeds are carried over (comments updated); all pass on the token path.
- Two intentional deltas, each with a test: `href` references inside comments, CDATA sections, processing instructions, or element text no longer refuse export (only real parsed attributes are inspected), and `data-href` keeps the historical conservative ends-with-`href` refusal. Root close matching is now case-consistent with root discovery (previously the close search was case-sensitive while discovery was not). An `href` attribute with no consumable value is refused by name, preserving fail-closed behavior for unparseable values.
- Mid-edit tolerance is preserved: a missing root close still slices to end-of-source, trailing garbage is ignored, and incomplete root tags still fail closed. No `SVGDocument` API change was needed.
- Scheme matching stays on raw value bytes (unchanged); entity-encoded schemes remain a known follow-up, not bundled here.

Supersedes #1308, which closes unmerged.

Verification:
- `viewport_svg_export_tests`: 57/57 pass, including the 6 carried tests (now with re-parse assertions), 2 behavior-delta tests, 2 nested-`svg` tests, and 1 unterminated-`href` refusal test.
- `xml_tests`: pass, including 4 new tokenizer convergence tests.
- `text_editor_tests`, `text_editor_core_tests`, `editor_widget_unit_tests` (44 autocomplete cases): pass.
- `tools/lint.sh`: clean. Changed files clang-format clean.
- Corpus replay of the 6 new seeds runs in the PR-gate fuzzer lane; local ASan replay was blocked by local disk exhaustion (environmental), so sanitizer coverage comes from CI.


jwmcglynn:
Superseded by #1314: the exporter reuses the parsed tree instead of the token stream. The 6 unit tests, 6 corpus seeds, and round-trip oracle from this PR are carried over there. Closing unmerged.

PR #1315, opened 2026-09-20
Silence warnings from the default opt editor build

jwmcglynn:
Cleans up the remaining output from a from-scratch `bazel run //donner/editor -c opt`: one root-module warning, two GeodeDevice compiler warnings, and two woff2 warnings.

- `GeodeDevice.cc`: the outstanding device-lost counters are only read by `assert(previous > 0)`, which compiles away under NDEBUG, so mark them `[[maybe_unused]]`. The atomic decrement and the debug assert are unchanged.
- `MODULE.bazel`: declare `rules_cc` 0.2.25 to match the resolved dependency graph (`apple_support` 2.8.2 already requires it). The CMake FetchContent mapping in `tools/cmake/gen_cmakelists.py` and both example modules are updated to the same version and commit.
- `third_party/BUILD.woff2`: add `-Wno-unused-but-set-variable` next to the existing woff2 warning suppressions for two dead stores in the vendored encoder.
- `third_party/bazel/non_bcr_deps.bzl`: per review, harfbuzz and tracy keep their readable release tags (`14.2.1`, `v0.13.1`) rather than being pinned to the raw commits their tags resolve to. Bazel still prints its canonical-form `DEBUG` hint for these two tagged git repos; that is accepted here rather than trading the tags for raw SHAs. No content change results: the module lockfile and the tags resolve to the same commits as before.

Verification:

- `bazel build -c opt //donner/editor:editor` succeeds with no `warning`/`WARNING` lines from Donner-owned or vendored code; woff2 and `GeodeDevice.cc` recompiled. The only remaining `DEBUG:` line is Bazel's canonical-form hint for the two tagged git repos, which is expected and not a compiler diagnostic.
- `bazel test //third_party/google_fonts/...`: 15/15 pass; the catalog codec fingerprint covers the overlay BUILD file, so this exercises the woff2 change.
- `bazel test //tools/cmake:gen_cmakelists_test` passes and `python3 tools/cmake/gen_cmakelists.py --check` passes.
- `tools/lint.sh`, `clang-format --dry-run -Werror`, and `buildifier -mode=check` on the touched Bazel files are clean.
- Independent non-implementing review approved the exact diff with no blocking findings; the security/privacy review found no issues.

No intentional behavior change.


third_party/bazel/non_bcr_deps.bzl
```diff
@@ -57,8 +57,8 @@ def _non_bcr_deps_impl(_mctx):
     new_git_repository(
         name = "harfbuzz",
         build_file = "//third_party:BUILD.harfbuzz",
+        commit = "56feae4035bdd48f62ba2b8d8c16232d4d89b3a4",  # tag 14.2.1
```
jwmcglynn: Keep the tag for both of these, don't link the exact commit

jwmcglynn: Kept the tags: reverted both to `tag = "14.2.1"` and `tag = "v0.13.1"` in 73b156524. No content change results because the tags resolve to the same commits that were pinned, and the module lockfile records the same attributes. Bazel's canonical-form DEBUG hint for these two repos remains, which the PR description now notes explicitly.

